PR #16, opened 2020-07-02
Add gpTapply method into GreenplumPython

HaozhouWang:
1. Add gpTapply methods
2. support extra args in gpTapply with pattern `type=vaule`
3. Refactor code
4. Add UT
5. run gpTapply without create table does not support yet
6. Distributed by does not support yet


HaozhouWang:
runtimeid and language does not support yet

PR #17, opened 2020-07-03
add pipeline

violet2016:


greenplumpython/gpapply.py
```diff
@@ -2,6 +2,11 @@
 import random
 import string
 import numpy as np
+<<<<<<< HEAD
+=======
+
+numpyDTypeToGPDBType = {np.dtype('int64'): "int", np.dtype('object'): "text", np.dtype("datetime64[ns]"):"datetime"}
+>>>>>>> change gpapply to run with new database & dataframe
```
HaozhouWang, line 9: conflict here

HaozhouWang:
we can change the file name and remove  `plcontainer`

greenplumpython/gpapply.py
violet2016: Thanks, fixed

PR #21, opened 2020-07-09
support runtime_id and runtime_type for plcontainer

wengyanqing:


HaozhouWang:
gpTapply will also need the supporting

wengyanqing:
> gpTapply will also need the supporting

Sure, fixed.

PR #29, opened 2020-07-22
Update setup version and script file

HaozhouWang:
1. Add versioneer package to get version from git tag
2. Update script file to collect all required files


versioneer.py
```diff
@@ -0,0 +1,1822 @@
...
+
+"""The Versioneer - like a rocketeer, but for versions.
+
+The Versioneer
+==============
+
+* like a rocketeer, but for versions!
+* https://github.com/warner/python-versioneer
+* Brian Warner
+* License: Public Domain
```
wengyanqing, line 12: Is this License compatible with greenplumpython?

HaozhouWang: Yes, it is compatible with apache license
https://creativecommons.org/publicdomain/zero/1.0/

PR #36, opened 2022-04-15
Make GreenplumPython work

Quasimewdo:
- Instead of using private function of sqlAlchemy, use
  public function in GPConnection()
- Connection url "pygresql" deprecated, change to "psycopg2"
- Load extra_args values in function body instead of call
  by function
- Index column can't be aggregate in gptApply()
  --> no aggregate on it
- Fixed tests

### DIFF
--- a/greenplumpython/core/gp_connection.py
+++ b/greenplumpython/core/gp_connection.py
@@ -16,18 +16,21 @@ class GPConnection(object):
         return conn_alchemy
 
     def connect(self, host: str, port: int, database: str, user: str, password: str):
-        url = "postgresql+pygresql://%s:%s@%s:%d/%s" % (user, password, host, port, database)
+        url = "postgresql+psycopg2://%s:%s@%s:%d/%s" % (user, password, host, port, database)
         engine = sqlalchemy.create_engine(url)
         conn = engine.connect()
-        conn_alchemy = self._sqlalchemy_connection_convert(conn)
-        conn_pairs = (conn, conn_alchemy)
-        self.connection_pool[self.max_connectid] = conn_pairs
-        self.max_connectid += 1 
-        return conn_alchemy
-    
+        #conn_alchemy = self._sqlalchemy_connection_convert(conn)
+        #conn_pairs = (conn, conn_alchemy)
+        #self.connection_pool[self.max_connectid] = conn_pairs
+        self.connection_pool[self.max_connectid] = conn
+        self.max_connectid += 1
+        #return conn_alchemy
+        return conn
+        
     def get_connection(self, conn_id: int):
         if conn_id in self.connection_pool.keys():
-            return self.connection_pool[conn_id][1]
+            #return self.connection_pool[conn_id][1]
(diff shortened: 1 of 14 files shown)


greenplumpython/core/gp_connection.py
```diff
@@ -16,18 +16,21 @@ def _sqlalchemy_connection_convert(self, conn):
...
-        conn_alchemy = self._sqlalchemy_connection_convert(conn)
-        conn_pairs = (conn, conn_alchemy)
-        self.connection_pool[self.max_connectid] = conn_pairs
-        self.max_connectid += 1 
-        return conn_alchemy
-    
+        #conn_alchemy = self._sqlalchemy_connection_convert(conn)
+        #conn_pairs = (conn, conn_alchemy)
+        #self.connection_pool[self.max_connectid] = conn_pairs
+        self.connection_pool[self.max_connectid] = conn
```
yihong0618, line 25: comments needs to delete when this `PR` done

greenplumpython/myexample.py
```diff
@@ -0,0 +1,61 @@
...
+username = "gpadmin"
+password = "changeme"
+conn = connection.connect(dbIP, dbPort, dbname, username, password)
+
+# init GPDB Database instance
+gpdb_instance = gp.GPDatabase(conn)
+
+# a dataframe warpper for input table
+schema = "public"
+table_name = "t2"
```
yihong0618, line 20: we need to add the create `t2` sql in case that all env can run this example

greenplumpython/tests/testdb.py
```diff
@@ -1,16 +1,17 @@
...
 db = os.getenv('GPDATABASE')
 if db is None or db is '':
-    db = "gppython"
+    db = "postgres"
 user = os.getenv('GPUSER')
 if user is None or user is '':
     user = "gpadmin"
-password = os.getenv('GPPASSWORD')
+#password = os.getenv('GPPASSWORD')
+password = "changeme"
```
yihong0618, line 17: need to delete

yihong0618 (commented):
seems that we need passed all the tests

greenplumpython/core/gp_connection.py
```diff
@@ -16,18 +16,21 @@ def _sqlalchemy_connection_convert(self, conn):
...
         conn = engine.connect()
-        conn_alchemy = self._sqlalchemy_connection_convert(conn)
-        conn_pairs = (conn, conn_alchemy)
-        self.connection_pool[self.max_connectid] = conn_pairs
-        self.max_connectid += 1 
-        return conn_alchemy
-    
+        #conn_alchemy = self._sqlalchemy_connection_convert(conn)
+        #conn_pairs = (conn, conn_alchemy)
+        #self.connection_pool[self.max_connectid] = conn_pairs
```
yihong0618, line 24: when merge we can delete these comments

Quasimewdo: Yes, I didn't delete them in case they do something special that we didn't identify yet.

PR #38, opened 2022-05-10
A pandas-like table API

xuebinsu:
More details will be added later.

### DIFF
--- a/greenplumpython/table.py
+++ b/greenplumpython/table.py
@@ -1,6 +1,7 @@
 import random
 import string
 from typing import Iterable, NamedTuple
+
 import sqlalchemy
 
 


greenplumpython/table.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        Returns
+        - a Column of the current Table if key is string, or
+        - a new Table from the current Table per the type of key:
+            - if key is a list, then SELECT a subset of columns, a.k.a. targets;
+            - if key is an Expr, then SELECT a subset of rows per the value of the Expr;
+            - if key is a slice, then SELECT a portion of consecutive rows
+        """
+        if isinstance(key, str):
+            return Column(key, self)
```
beeender, line 58: So the column will be a special `Table` which contains one column only?

greenplumpython/table.py
```diff
@@ -0,0 +1,140 @@
...
+                f"SELECT * FROM {self.name} LIMIT {key.stop} {offset_clause}",
+                parents=[self],
+            )
+
+    def list_lineage(self) -> Iterable["Table"]:
+        lineage = [self]
+        current = 0
+        while current < len(lineage):
+            for table in lineage[current].parents:
+                lineage.append(table)
```
yihong0618, line 80: is logic here len(lineage) always==1?

xuebinsu: No. More items will be appended to the list as the loop goes.

yihong0618: cool missed thanks

greenplumpython/table.py
```diff
@@ -0,0 +1,140 @@
...
+    ) -> None:
+        self.query = query
+        self.parents = parents
+        if table_name is not None:
+            self.name = table_name
+        else:
+            self.name = "tb_" + "".join(
+                random.choice(string.ascii_lowercase) for _ in range(60)
+            )
+        if any(parents):
```
yihong0618, line 43: cool

yihong0618 (approved):
LGTM

PR #45, opened 2022-06-01
Support creating UDFs from Python functions

xuebinsu:
It is common that a data scientist needs to define new UDFs for
complex analytics tasks. This patch introduces a new decorator that
converts a Python function into a Greenplum UDF. And the created
UDF can be called in exactly the same way as ordinary Python
functions. This makes it much easier to work with Greenplum UDFs.


greenplumpython/type.py
```diff
@@ -0,0 +1,14 @@
...
+    int: "integer",
+    float: "double precision",
+    bool: "boolean",
+    str: "text",
+    bytes: "bytea",
+}
+
+
+def is_primitive(type_: Type) -> bool:
+    return type_ in primitive_type_map
```
yihong0618, line 14: is this functions used?

yihong0618 (commented):
cool

Quasimewdo (approved):

greenplumpython/table.py
```diff
@@ -101,7 +101,7 @@ def fetch(self, all: bool = True) -> Iterable:
         return self._db.execute(self._build_full_query())
 
     def save_as(
-        self, table_name: str, temp: bool = True, column_names: Iterable[str] = []
+        self, table_name: str, temp: bool = False, column_names: Iterable[str] = []
```
beeender, line 104: Why this default value is changed?

tests/test_func.py
```diff
@@ -20,3 +22,13 @@ def test_set_returning_func(db: gp.Database):
...
+
+
+def test_create_func(db: gp.Database):
+    @gp.create_function(db)
+    def add(a: int, b: int) -> int:
+        return a + b
+
+    for row in add(1, 2, as_name="sum").to_table().fetch():
+        assert row["sum"] == 1 + 2
+        assert row["sum"] == inspect.unwrap(add)(1, 2)
```
beeender, line 34: Add test `create_function()` with no param type in the signature.
```
@gp.create_function(db)
    def add(a, b) -> int:
        return a + b
```

Test with multi-lines function body to make sure the indention is correct

```
@gp.create_function(db)
    def add(a, b) -> int:
        c = a + b
        return c
```

Test with space/tab indention function body 


`# fmt: on/off` may help if you want to turn the black format off.

greenplumpython/func.py
```diff
@@ -26,11 +30,54 @@ def to_table(self) -> Table:
...
         return FunctionCall(name, db, args, as_name=as_name)
 
-    return make_callable
+    return make_function_call
 
 
-def create_function():
-    pass
+def create_function(
+    db: Database,
```
beeender, line 40: I am a bit worried about the db param here.
By specifying the `db` in the `create_function()`, the user has to have a `db` instance before declaring the UDF. And if user has some common UDFs across different database, do they have to start a new greenplumpython process? If that is the case, our `Database` class should just be a singleton and we don't need `db` for each function.

I am thinking about delaying the UDF declaration until the execution when we know which db to use.

I can see it will be some amount of work, we can discuss and implement it in other PRs if needed.

beeender: I seriously think the UDF creation should be delayed until it is needed. We CAN NOT ask the user to have a valid gpdb connection when she/he is writing and testing the UDF body in its local python env (not the gpdb plpython).

tests/test_func.py
xuebinsu: Currently, the `db` param is required.

The indention issue is fixed.

greenplumpython/table.py
xuebinsu: Because I realized that if I want to save some data to the database, I would expect that the data is persistent after my work is finished. 

But if I want to create a function, I would expect that the function is deleted as soon as I finish using it.

greenplumpython/func.py
xuebinsu: Changed.

tests/test_func.py
xuebinsu: Added a test case for tab-indented function definition

Added a TODO for testing

greenplumpython/table.py
beeender: Interesting!

tests/test_func.py
```diff
@@ -20,3 +22,40 @@ def test_set_returning_func(db: gp.Database):
...
+        if a > b:
+            return a
+        else:
+            return b
+
+    for row in my_max(1, 2, as_name="result", db=db).to_table().fetch():
+        assert row["result"] == max(1, 2)
+        assert row["result"] == inspect.unwrap(my_max)(1, 2)
+
+# fmt: on/off
```
beeender, line 50: try:
```
# fmt: off
def test_xxx
    xxxx
# fmt: on
```

tests/test_func.py
xuebinsu: create_function() with no param type currently does NOT work.

tests/test_func.py
xuebinsu: Done.

tests/test_func.py
xuebinsu: Maybe we can add an optional `signature` argument to allow the user specify the signature explictly.

xuebinsu:
> Can we pass columns as UDF's arguments? Maybe for later

@Quasimewdo This is exactly the next step.

beeender (approved):
:+1:  If CI is happy

PR #47, opened 2022-06-01
Fix lint errors reported by pyright

Quasimewdo:
- Add NodeJs version requirement in README.dev.md
- TODO : Fix Lint errors and remove ignore list

### DIFF
--- a/README.dev.md
+++ b/README.dev.md
@@ -16,6 +16,10 @@ Install with `brew` on macOS:
 brew install tox
 ```
 
+`NodeJs` need to have version 12.X, otherwise `Pyright` can't be executed correctly. Possible scenarios:
+ - Error : Cannot find module 'worker_threads'
+ - Nothing happens when run `Pyright`
+
 ## Build & Test
 
 ### Lint


greenplumpython/db.py
```diff
@@ -11,7 +11,9 @@ def __init__(self, **params) -> None:
             cursor_factory=psycopg2.extras.RealDictCursor,
         )
 
-    def execute(self, query: str, args: Iterable = [], has_results: bool = True) -> Iterable:
+    def execute(
+        self, query: str, args: Optional[List[str]] = None, has_results: bool = True
```
beeender, line 15: `args` doesn't have to be `Optional`, `List` should be enough

```
    def execute(self, query: str, args: List = [], has_results: bool = True) -> Optional[Iterable]:
        with self._conn.cursor() as cursor:
            cursor.execute(query, args)
            return cursor.fetchall() if has_results else None
```
Would the above be enough to fix the lint issue?

Why does `args` need to be `List[str]`? Actually, I don't think we need the `args` param, do we? @xuebinsu

greenplumpython/table.py
```diff
@@ -69,18 +69,23 @@ def name(self) -> str:
...
 
     @property
-    def db(self) -> db.Database:
+    def db(self) -> Optional[db.Database]:
         return self._db
 
-    def _list_lineage(self) -> Iterable["Table"]:
-        lineage = [self]
+    @property
+    def parents(self) -> Iterable["Table"]:
```
beeender, line 76: Should `parents` be exposed? Is this related to the lint error?

greenplumpython/db.py
xuebinsu: This depends on whether the database driver exposes the function to convert a Python constant, e.g. a byte array, to a Postgres constant.
- If yes, then we do not need this argument, anything can be converted to a string.
- If no, then we need to use the query string as only the template and pass the arguments separately.

greenplumpython/table.py
```diff
@@ -118,7 +125,7 @@ def save_as(
 
 
 # table_name can be table/view name
-def table(name: str, db: db.Database) -> Table:
+def table(name: str, db: Optional[db.Database]) -> Table:
```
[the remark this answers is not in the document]
xuebinsu: I don't think so. We can raise exceptions when `db` is `None` in the `Table` class to pass the type checking.

greenplumpython/table.py
```diff
@@ -94,19 +99,21 @@ def _build_full_query(self) -> str:
                 cte_list.append(f"{table._name} AS ({table._query})")
         return "WITH " + ",".join(cte_list) + self._query
 
-    def fetch(self, all: bool = True) -> Iterable:
+    def fetch(self, all: bool = True) -> Optional[List]:
```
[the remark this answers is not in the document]
xuebinsu: I agree. Besides, when `all` is `False`, the result is not a `list` but only an `Iterable`.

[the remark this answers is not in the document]
Quasimewdo: `        result = self._db.execute(self._build_full_query())
         return result if result is not None else []`

greenplumpython/table.py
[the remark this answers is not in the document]
Quasimewdo:
```suggestion
def table(name: str, db: db.Database) -> Table:
```

greenplumpython/db.py
Quasimewdo:
```suggestion
            def execute(self, query: str, args: List = [], has_results: bool = True) -> Optional[Iterable]:
```

greenplumpython/table.py
```diff
@@ -102,11 +103,14 @@ def fetch(self, all: bool = True) -> Iterable:
         """
         if not all:
             raise NotImplementedError()
-        return self._db.execute(self._build_full_query())
+        assert self._db is not None
+        result = self._db.execute(self._build_full_query())
+        return result if result is not None else []
```
beeender, line 108: The `result` should not be `None` here IMO.

```
        assert result is not None
        return result
```

What about this?

beeender (approved):
With one comment

PR #48, opened 2022-06-02
Add plpython3 to concourse

beeender:
So we can use plpython3u in the test pipeline.

### DIFF
--- a/concourse/scripts/entry.sh
+++ b/concourse/scripts/entry.sh
@@ -114,6 +114,7 @@ function install_gpdb() {
     # Start cluster
     source "/home/gpadmin/gpdb_src/concourse/scripts/common.bash"
     make_cluster
+    source /home/gpadmin/gpdb_src/gpAux/gpdemo/gpdemo-env.sh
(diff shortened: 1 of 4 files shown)


xuebinsu (approved):

concourse/scripts/entry.sh
```diff
@@ -124,10 +125,15 @@ function setup_gpadmin_bashrc() {
     } >> /home/gpadmin/.bashrc
 }
 
+function install_plpython3() {
+    PGUSER=gpadmin gppkg -i bin_plpython3/*.gppkg
```
yihong0618, line 129: why we need to add `PGUSER=gpadmin` prefix here?

yihong0618 (approved):
LGTM

concourse/scripts/entry.sh
beeender: gppkg needs to run sql and by default, it uses the current user which is root. But our gpdb setup doesn't have a root user.

PR #49, opened 2022-06-02
Support Table Join:

Quasimewdo:
- Add join function in table.py
    * Consider different cases of join
    * Consider target_lists
- Add query test for binary_operator between two columns
- Add test for join in test_const_table.py
- Add parameter temp = True for test table
- TODO : Add extra arguments to rename columns


greenplumpython/table.py
```diff
@@ -52,6 +52,26 @@ def select(self, target_list: Iterable) -> "Table":
             parents=[self],
         )
 
+    def join(
+        self,
+        other: "Table",
+        expr: expr.Expr,
```
xuebinsu, line 58: 1. `expr` is too general here. Better to use a more specific name such as `cond` for "condition"
2. `expr` should be optional. `CROSS JOIN` does not require any condition.

greenplumpython/table.py
```diff
@@ -52,6 +52,26 @@ def select(self, target_list: Iterable) -> "Table":
...
+        how: Optional[str] = "natural",
```
xuebinsu, line 59: I think it is more clear to support multiple JOIN types with different methods, e.g., `outer_join()`, `cross_join()`, etc., insead of using the `how` argument here.

greenplumpython/table.py
Quasimewdo:
```suggestion
        Optional[cond]: expr.Expr,
```

greenplumpython/table.py
```diff
@@ -52,6 +52,26 @@ def select(self, target_list: Iterable) -> "Table":
...
+        target_list: Optional[Iterable] = None,
+    ) -> "Table":
+        select_str = (
+            ",".join([str(target) for target in target_list]) if target_list is not None else "*"
+        )
+        on_str = "ON " + str(expr) if how.upper() != "NATURAL" else None
```
yihong0618, line 65: this `on_str` can be `None`
it this fit for
```
                SELECT {select_str} 
                FROM {self.name} {how.upper()} JOIN {other.name}
                {str(on_str)}  
```
str(None) == 'None'

greenplumpython/table.py
Quasimewdo: If we separate them don't we have repeated lines of code?

xuebinsu: We can have a private function that do the real job, and the public functions do nothing but call the private function

Quasimewdo: Got it

greenplumpython/table.py
Quasimewdo: Because when we have `Natural Join` then `ON` statement is not required

tests/test_const_table.py
```diff
@@ -25,6 +25,66 @@ def test_const_table(db: gp.Database):
...
+
+
+def test_table_self_join(db: gp.Database):
+    # fmt: off
+    rows1 = [(1, "'a1'",), (2, "'a2'",), (3, "'a3'",)]
+    # fmt: on
+    t1 = gp.values(rows1, db=db).save_as("temp1", temp=True, column_names=["id1", "n1"])
+    t2 = t1.as_name("temp2")
+    ret = t1.inner_join(t2, t1["id1"] == t2["id1"]).fetch()
+    assert len(list(ret)) == 3
```
xuebinsu, line 90: 1. Need to check what columns are in the results using `assert`. The columns will be referenced later by the user.
2. Need to test whether the resulting table can be saved.

xuebinsu (changes requested):

xuebinsu (approved):

tests/test_const_table.py
```diff
@@ -25,6 +25,97 @@ def test_const_table(db: gp.Database):
...
+    # fmt: off
+    rows1 = [(1, "'a1'",), (2, "'a2'",), (3, "'a3'",)]
+    # fmt: on
+    t1 = gp.values(rows1, db=db).save_as("temp1", temp=True, column_names=["id1", "n1"])
+    t2 = t1.as_name("temp2")
+    t_join = t1.inner_join(
+        t2,
+        t1["id1"] == t2["id1"],
+        targets1=["*"],
+        targets2=["id1 AS t2_id1", "n1 AS t2_n1"],
```
xuebinsu, line 114: This is still part of SQL. Probably need to change this.

greenplumpython/table.py
```diff
@@ -52,6 +55,85 @@ def select(self, target_list: Iterable) -> "Table":
...
+        targets1: List = ["*"],
+        targets2: List = ["*"],
+    ):
+        on_str = " ".join(["ON", str(cond)])
+        return self._join(other, targets1, targets2, "RIGHT JOIN", on_str)
+
+    def full_outer_join(
+        self,
+        other: "Table",
+        cond: Optional[expr.Expr] = None,
```
beeender, line 112: ditto

tests/test_const_table.py
```diff
@@ -25,6 +25,97 @@ def test_const_table(db: gp.Database):
 def test_table_getitem(db: gp.Database):
     rows = [(1,), (2,), (3,)]
     t = gp.values(rows, db=db)
-    t = t.save_as("const_table", column_names=["id"])
+    t = t.save_as("const_table", temp=True, column_names=["id"])
     c = t["id"]
     assert str(c) == (t.name + ".id")
+
+
```
beeender, line 32: `join` is a big and complex feature which worth a separate test file. I propose to create a new file `test_join.py` and organize the test like the below:

**Tests on `_join` method.**
Since it is the base method for all the "join" APIs, we could use those tests to verify common things like the query result columns based on different input combinations. In those tests, the output tuples are not important, and tuples verification should be coved by other tests.

**Tested on individual public "join" method.**
For those tests, the result tuples actually matter. And since different joins may have different characteristics, the tests should cover all the differences of those.


"join" could be complex, to prevent this from becoming a huge PR, choosing a few public methods to implement as a start might be a good idea. But that totally depends on you.

greenplumpython/table.py
```diff
@@ -52,6 +55,85 @@ def select(self, target_list: Iterable) -> "Table":
             parents=[self],
         )
 
+    def _join(
+        self,
+        other: "Table",
+        targets1: List = ["*"],
+        targets2: List = ["*"],
+        how: Optional[str] = "NATURAL JOIN",
+        on_str: Optional[str] = None,
```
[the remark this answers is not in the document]
Quasimewdo:
```suggestion
        on_str: str = "",
```

greenplumpython/table.py
```diff
@@ -52,6 +55,85 @@ def select(self, target_list: Iterable) -> "Table":
             parents=[self],
         )
 
+    def _join(
+        self,
+        other: "Table",
+        targets1: List = ["*"],
+        targets2: List = ["*"],
+        how: Optional[str] = "NATURAL JOIN",
```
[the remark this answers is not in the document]
Quasimewdo:
```suggestion
        how: str = "JOIN",
```

greenplumpython/table.py
```diff
@@ -52,6 +55,85 @@ def select(self, target_list: Iterable) -> "Table":
             parents=[self],
         )
 
+    def _join(
+        self,
+        other: "Table",
+        targets1: List = ["*"],
+        targets2: List = ["*"],
```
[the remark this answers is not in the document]
Quasimewdo: > we should have tests on the `_join()` to cover most of the scenarios of different input targets. Since other public methods are simple wrappers of the `_join()`, that would be enough. In the tests cases, we need to test:
> 
> 1. both targets have default values. (`SELECT *, * from t1 join t2 on t1.id = t2.id`, this would have questionable behavior, why does the user ever need two `*` ?)
> 2. both targets have empty targets
> 3. One empty, one default
> 4. both have multiple targets
> 5. non-exist target, do we have a clear error?
> 6. other necessary combinations.
> 
> The tests will focus on the columns, so we don't have to test the tuples to simplify the cases.
> 
> If we are uncertain about some cases, leave a FIXME in the code so we can fix them later.

1.To handle both `*` : in order to manage same name columns, rename columns which have identical name than others in format "TableName_ColumnName'  
2. If both targets lists empty, same as two `*`
5. Raise error if column not exists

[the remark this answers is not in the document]
xuebinsu: I am thinking about having something like
```python
t1.join(t2, targets=[t1["id"].rename_to("t1_id"), t2["id"].rename_to("t2_id")])
```
where `rename_to()` is a method of an `Expr`:
```python
def rename_to(self, new_name: str) -> Expr:
    new_expr = copy.copy(self) # Shallow copy
    new_expr._as_name = new_name
    return new_expr
```

xuebinsu (approved):
LGTM

greenplumpython/expr.py
```diff
@@ -45,6 +46,14 @@ def __ne__(self, other):
...
 
     def __str__(self) -> str:
+        return self._mystr() + (" AS " + self._as_name if self._as_name is not None else "")
+
+    def rename(self, new_name: str) -> "Expr":
+        new_expr = copy.copy(self)  # Shallow copy
+        new_expr._as_name = new_name
+        return new_expr
+
+    def _mystr(self) -> str:
```
beeender, line 56: `_mystr` is really not a good name.
What about `_sql_str`? Or we can change it later when a better name comes to mind.

greenplumpython/table.py
```diff
@@ -52,6 +55,85 @@ def select(self, target_list: Iterable) -> "Table":
             parents=[self],
         )
 
+    def _join(
+        self,
+        other: "Table",
```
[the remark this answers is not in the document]
beeender: shall we do anything to this comment? I don't think we should give `targets`/`how`/`on_str` default values.
how about just
```
def _join(
    self,
    other: "Table",
    targets: List,
    how: str,
    on_str: str
) -> "Table"
```

greenplumpython/table.py
```diff
@@ -52,6 +55,77 @@ def select(self, target_list: Iterable) -> "Table":
...
+        on_str = " ".join(["ON", str(cond)])
+        return self._join(other, targets, "FULL JOIN", on_str)
+
+    def natural_join(
+        self,
+        other: "Table",
+        targets: List = [],
+    ):
+        on_str = ""
+        return self._join(other, targets, "FULL JOIN", on_str)
```
beeender, line 119: This is still wrong. `FULL JOIN` -> `NATURAL JOIN`. And you don't have right tests cover this.

It is strange that my old comments disappeared.
https://github.com/greenplum-db/GreenplumPython/pull/49/files/61eb1e0920fdf833eced6d2429965dd9266d9856#r894127757

tests/test_column.py
```diff
@@ -38,3 +38,7 @@ def test_expr_column_str_in_query(db: gp.Database, table: gp.Table):
         keys = list(row.keys())
         assert len(keys) == 1
         assert "id" in keys[0]
+
+
+def test_expr_column_rename(db: gp.Database, table: gp.Table):
```
beeender, line 43: We also need to verify that the `rename()` returns a new object instead of modifying the old one.
try below
```
t = table["id"]
t_renamed = t.rename("table_id")
assert t is not t_renamed
assert str(t_renamed) == "const_table.id AS table_id"
```

tests/test_join.py
```diff
@@ -0,0 +1,266 @@
...
+    ).fetch()
+    assert list(list(ret)[0].keys()) == ["id1", "idd1", "n1", "id2", "idd2", "n2"]
+
+
+def test_join_both_empty_targets(db: gp.Database, t1: gp.Table, t2: gp.Table):
+    ret = t1._join(t2, targets=[], how="INNER JOIN", on_str="ON temp1.id1 = temp2.id2").fetch()
+    assert list(list(ret)[0].keys()) == ["id1", "idd1", "n1", "id2", "idd2", "n2"]
+
+
+def test_join_all_emp_targets(db: gp.Database, t1: gp.Table, t2: gp.Table):
```
beeender, line 63: `test_join_all_emp_targets` what does `emp` mean here? typo?

greenplumpython/table.py
```diff
@@ -52,6 +55,85 @@ def select(self, target_list: Iterable) -> "Table":
...
+        return self._join(other, targets1, targets2, "FULL JOIN", on_str)
+
+    def natural_join(
+        self,
+        other: "Table",
+        targets1: List = ["*"],
+        targets2: List = ["*"],
+    ):
+        on_str = ""
+        return self._join(other, targets1, targets2, "FULL JOIN", on_str)
```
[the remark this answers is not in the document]
Quasimewdo: Didn't notice, sorry, will fix it

tests/test_join.py
Quasimewdo: > `test_join_all_emp_targets` what does `emp` mean here? typo?

One table with all columns and one table with empty targets list

tests/test_join.py
```diff
@@ -60,7 +61,7 @@ def test_join_both_empty_targets(db: gp.Database, t1: gp.Table, t2: gp.Table):
     assert list(list(ret)[0].keys()) == ["id1", "idd1", "n1", "id2", "idd2", "n2"]
 
 
-def test_join_all_emp_targets(db: gp.Database, t1: gp.Table, t2: gp.Table):
+def test_join_all_empty_targets(db: gp.Database, t1: gp.Table, t2: gp.Table):
```
beeender, line 64: I still don't get why the test is named as `xxx_empty_xxx` here, what does the empty mean here?

beeender (approved):
one minor comment. Looks good!

tests/test_join.py
Quasimewdo: > I still don't get why the test is named as `xxx_empty_xxx` here, what does the empty mean here?

It means for example we want to select all columns of table t1 and 0 columns of table t2. Well is not very meaningful in a join case but added this test case.

PR #50, opened 2022-06-06
Support function calls on table columns

xuebinsu:
Previously, we implemented function calls with no argument, e.g.
`version()`, and function calls on constants, e.g. `add(1, 2)`. This
patch adds support for function calls on table columns, e.g.
`SELECT multiply(t.a, t.b) FROM t`. This makes the pandas-like API
more powerful.

### DIFF
--- a/greenplumpython/expr.py
+++ b/greenplumpython/expr.py
@@ -83,6 +83,7 @@ class Column(Expr):
         self._name = name
 
     def __str__(self) -> str:
+        assert self.table is not None
         return self.table.name + "." + self.name
 
     @property
--- a/greenplumpython/expr.py
+++ b/greenplumpython/expr.py
@@ -7,9 +7,15 @@ if TYPE_CHECKING:
 
 
 class Expr:
-    def __init__(self, as_name: Optional[str] = None, db: Optional[Database] = None) -> None:
+    def __init__(
+        self,
+        as_name: Optional[str] = None,
+        table: Optional[Database] = None,
+        db: Optional[Database] = None,
+    ) -> None:
         self._as_name = as_name
-        self._db = db
+        self._table = table
+        self._db = table.db if table is not None else db
 
     def __eq__(self, other):
         if isinstance(other, type(None)):
@@ -42,6 +48,10 @@ class Expr:
     def db(self) -> Database:
         return self._db
 
+    @property
+    def table(self) -> "Table":
+        return self._table
+
 
 class BinaryExpr(Expr):
     def __init__(self, operator: str, left: Expr, right, as_name: Optional[str] = None):
@@ -66,8 +76,7 @@ class BinaryExpr(Expr):
 
 class Column(Expr):
     def __init__(self, name: str, table: "Table", as_name: Optional[str] = None) -> None:
-        super().__init__(as_name=as_name)
-        self._table = table
+        super().__init__(as_name=as_name, table=table)
         self._name = name
 
     def __str__(self) -> str:
--- a/greenplumpython/expr.py
+++ b/greenplumpython/expr.py
@@ -50,7 +50,7 @@ class Expr:
         return self._db
 
     @property
-    def table(self) -> "Table":
+    def table(self) -> Optional["Table"]:
         return self._table
 
 
@@ -88,7 +88,3 @@ class Column(Expr):
     @property
     def name(self) -> str:
         return self._name
-
-    @property
-    def table(self) -> "Table":
-        return self._table
--- a/greenplumpython/expr.py
+++ b/greenplumpython/expr.py
@@ -56,7 +56,9 @@ class Expr:
 
 class BinaryExpr(Expr):
     def __init__(self, operator: str, left: Expr, right, as_name: Optional[str] = None):
-        super().__init__(as_name=as_name)
+        table = left.table if left is not None and left.table is not None else right.table
+        db = left.db if left is not None and left.db is not None else right.db
+        super().__init__(as_name=as_name, table=table, db=db)
         self.operator = operator
         self.left = left
         self.right = right
--- a/greenplumpython/expr.py
+++ b/greenplumpython/expr.py
@@ -16,6 +16,7 @@ class Expr:
         self._as_name = as_name
         self._table = table
         self._db = table.db if table is not None else db
+        assert self._db is not None
 
     def __eq__(self, other):
         if isinstance(other, type(None)):
--- a/greenplumpython/expr.py
+++ b/greenplumpython/expr.py
@@ -10,7 +10,7 @@ class Expr:
     def __init__(
         self,
         as_name: Optional[str] = None,
-        table: Optional[Database] = None,
+        table: Optional["Table"] = None,
         db: Optional[Database] = None,
     ) -> None:
         self._as_name = as_name
--- a/greenplumpython/func.py
+++ b/greenplumpython/func.py
@@ -12,12 +12,23 @@ from .type import primitive_type_map
 
 class FunctionCall(Expr):
     def __init__(
-        self, func_name: str, db: Database, args: Iterable[Expr] = [], as_name: Optional[str] = None
+        self,
+        func_name: str,
+        args: Iterable[Expr] = [],
+        as_name: Optional[str] = None,
+        db: Optional[Database] = None,
     ) -> None:
-        super().__init__(as_name)
+        table: Optional[Table] = None
+        for arg in args:
+            if isinstance(arg, Expr) and arg.table is not None:
+                if table is None:
+                    table = arg.table
+                elif table.name != arg.table.name:
+                    raise Exception("Cannot pass arguments from more than one tables")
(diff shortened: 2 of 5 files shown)


greenplumpython/expr.py
```diff
@@ -7,9 +7,15 @@
 
 
 class Expr:
-    def __init__(self, as_name: Optional[str] = None, db: Optional[Database] = None) -> None:
+    def __init__(
+        self,
+        as_name: Optional[str] = None,
+        table: Optional["Table"] = None,
+        db: Optional[Database] = None,
```
Sasasu, line 14: I have a question:

In fact, we did not support multiple database in one SQL. and dblink need some special API to do cross database query.

why not put database in `fetch()`.

or some context API (or session API), the may API looks:

```py
gp.GlobalContext = gp.ContextBuilder().WithDatabase("xxxx");

"some very long SQL build".fetch();

or

"some very long SQL build".fetch(gp.GlobalContext);
```

xuebinsu: Great  question! Even though we cannot access multiple database in one SQL statement, I think it is perfectly valid if the user wants to use GreenplumPython as a "pipe" to transmit data between any two databases. For example, 
```python
t1 = gp.table("t1", db=db1)
data = t1.fetch()
t2 = gp.table("t2", db=db2)
t2.insert(data)
```

Sasasu: > Great question! Even though we cannot access multiple database in one SQL statement, I think it is perfectly valid if the user wants to use GreenplumPython as a "pipe" to transmit data between any two databases. For example,

in session API:

```python
data = gp.table("t1").fetch(gp.ContextBuilder().WithDatabase("db1"))
gp.table("t2", db=db2).insert().execute(gp.ContextBuilder().WithDatabase("db2"))
```

xuebinsu: Why do you want to introduce this `ContextBuilder()`?

Sasasu: > Why do you want to introduce this `ContextBuilder()`?

not introduce `ContextBuilder`, think some global connection will be better, the user doesn't need to remember which table has a database and which table don't.

the API can be `ContextBuilder()` or magic string (`%sql`) in ipyhon, or `SparkSession` in pyspark 
```python
from pyspark.sql import SparkSession
spark = SparkSession.builder.master("local[1]").appName("SparkByExamples.com").getOrCreate() 
```

Sasasu: and the hierarchy, current:

```
Expr         = Table      + Database
BinaryExpr   = Expr       + Operator   + Expr
FunctionCall = "Function" + Database
Column       = "Column"   + Table
Table        = "Query"    + [Table]    + "Name" + Database
```

the hierarchy does not very clear, just simple and works.
I think we should follow the SQL AST to order our hierarchy.

for example, in `SELECT` `DISTINCT` is a keyword, not an expression (or an operator)

```
[ WITH [ RECURSIVE ] with_query [, ...] ]
SELECT [ ALL | DISTINCT [ ON ( expression [, ...] ) ] ]
    [ * | expression [ [ AS ] output_name ] [, ...] ]
    [ FROM from_item [, ...] ]
    [ WHERE condition ]
    [ GROUP BY [ ALL | DISTINCT ] grouping_element [, ...] ]
    [ HAVING condition ]
    [ WINDOW window_name AS ( window_definition ) [, ...] ]
    [ { UNION | INTERSECT | EXCEPT } [ ALL | DISTINCT ] select ]
    [ ORDER BY expression [ ASC | DESC | USING operator ] [ NULLS { FIRST | LAST } ] [, ...] ]
    [ LIMIT { count | ALL } ]
    [ OFFSET start [ ROW | ROWS ] ]
    [ FETCH { FIRST | NEXT } [ count ] { ROW | ROWS } { ONLY | WITH TIES } ]
    [ FOR { UPDATE | NO KEY UPDATE | SHARE | KEY SHARE } [ OF table_name [, ...] ] [ NOWAIT | SKIP LOCKED ] [...] ]
```
it means we can not add multiple `DISTINCT` in one SELECT part, `DISTINCT` is a part of `SelectStmt`.

so the API for `SelectStmt` looks like:

```
.all(exp: [Expression])                   -> SelectStmt(without all and distinct )
.distinct(exp: [Expression])              -> SelectStmt(without all and distinct )
.from([exp: Expression, as: Str])         -> SelectStmt(without from)
.where(con: Condition)                    -> SelectStmt(without where)
.GroupBy(grouping_element: [str])         -> SelectStmt(without groupby groupbyall groupbydistinct)
.GroupByAll(grouping_element: [str])      -> SelectStmt // pg9 not support this
.GroupByDistinct(grouping_element: [str]) -> SelectStmt // pg9 not support this
.Having(con: condition)                   -> SelectStmt(whiout having)

and more things like that format

```

the final thing our user needs to do does not change with this wrap.

```
gp.table(tablename: str)     -> from_item // a subtype of table, same as new_table in InsertStmt
from_item.fetch()            -> Iter      // wrap with select * from <table>
expression.fetch()           -> Iter      // wrap with select * from <expression>
from_item.insert(expression) -> void      // insert into <from_item> values <expression>

user can still do:

t1 = gp.table("t1", db=db1)
data = t1.fetch()
t2 = gp.table("t2", db=db2)
t2.insert(data)
```

here is PostgreSQL's expression type https://www.postgresql.org/docs/current/sql-expressions.html#SQL-EXPRESSIONS-COLUMN-REFS
and here is AST for SELECT https://www.postgresql.org/docs/9.6/sql-select.html
also AST for INSERT https://www.postgresql.org/docs/9.6/sql-insert.html

greenplumpython/func.py
```diff
@@ -12,26 +12,43 @@
...
         as_string = f"AS {self._as_name}" if self._as_name is not None else ""
-        ret_table = Table(f"SELECT * FROM {str(self)} {as_string}", db=self._db)
-        return Table(f"SELECT * FROM {ret_table.name}", parents=[ret_table], db=self._db)
+        from_caluse = f"FROM {self.table.name}" if self.table is not None else ""
+        parents = [self.table] if self.table is not None else []
+        orig_func_table = Table(
+            f"SELECT {str(self)} {as_string} {from_caluse}", db=self._db, parents=parents
+        )
+        return Table(
+            f"SELECT * FROM {orig_func_table.name}", parents=[orig_func_table], db=self._db
```
Sasasu, line 45: assuming a `orig_func_table` is can be `from_item` in AST type.
what if the user does insert on this table?

this comment not for this PR, I think we should fix this before GreenplumPython becomes complex.

Sasasu (approved):

greenplumpython/func.py
```diff
@@ -12,26 +12,43 @@
...
     ) -> None:
-        super().__init__(as_name)
+        table: Optional[Table] = None
+        for arg in args:
+            if isinstance(arg, Expr) and arg.table is not None:
+                if table is None:
+                    table = arg.table
+                elif table.name != arg.table.name:
+                    raise Exception("Cannot pass arguments from more than one tables")
+        db = table.db if table is not None else db
```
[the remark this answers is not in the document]
Quasimewdo: > this line is not needed if you have expr.py line 18.

But there is no super() call to use this line

Quasimewdo (commented):
LGTM

greenplumpython/expr.py
xuebinsu: Thanks for your feedback! Will consider these issues.

greenplumpython/func.py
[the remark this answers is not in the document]
xuebinsu:
```suggestion
```

greenplumpython/expr.py
```diff
@@ -7,9 +7,15 @@
...
+    ) -> None:
         self._as_name = as_name
-        self._db = db
+        self._table = table
+        self._db = table.db if table is not None else db
```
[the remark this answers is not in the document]
xuebinsu:
```suggestion
        self._db = table.db if table is not None else db
        assert self._db is not None
```

greenplumpython/func.py
xuebinsu: Thanks for your feedback! Will consider this.

tests/test_func.py
```diff
@@ -62,3 +61,22 @@ def my_min(a: int, b: int) -> int:
...
+
+def test_func_on_multi_columns(db: gp.Database):
+    @gp.create_function
+    def multiply(a: int, b: int) -> int:
+        return a * b
+
+    rows = [(i, i) for i in range(10)]
+    series = gp.values(rows, db=db, column_names=["a", "b"])
+    results = multiply(series["a"], series["b"], as_name="result").to_table().fetch()
+    assert sorted([row["result"] for row in results]) == [i * i for i in range(10)]
```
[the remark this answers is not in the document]
xuebinsu: Test case added.

Quasimewdo (approved):
LGTM

beeender (approved):

PR #53, opened 2022-06-09
Support get table on slice

Quasimewdo:
During data analysis, users may want to select a slice of the table.
This patch supports slice selection without order on Greenplum
tables using "[]", and a function table.top() which supports a 
slice selection with an order defined by user.

### DIFF
--- a/greenplumpython/table.py
+++ b/greenplumpython/table.py
@@ -66,8 +66,21 @@ class Table:
             parents=[self],
         )
 
+    @staticmethod
+    def _order_by_str(order_by: Iterable) -> str:
+        order_by_clause = (
+            f"""
+                {','.join([' '.join([col, order]) for col, order in order_by.items()])}
+            """
+            if isinstance(order_by, dict)
+            else f"""
+                    {','.join([order_index for order_index in order_by])}
+                """
+        )
+        return order_by_clause
+
     def top(self, count: int, order_by: Iterable, skip: int = 0):
-        order_by_clause = order_by_str(order_by)
+        order_by_clause = self._order_by_str(order_by)
         return Table(
             f"""
                 SELECT * FROM {self.name}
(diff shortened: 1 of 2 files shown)


greenplumpython/table.py
```diff
@@ -11,10 +11,12 @@ def __init__(
         parents: Iterable["Table"] = [],
         name: Optional[str] = None,
         db: Optional[db.Database] = None,
+        order_by: Optional[Union[Iterable, dict]] = None,
```
beeender, line 14: when handle `group by`, it was put into the query string and created a new table

https://github.com/greenplum-db/GreenplumPython/blob/cf0ce562390836f17b656645770e83156b78ae16/greenplumpython/func.py#L51

```
        orig_func_table = Table(
            " ".join(
                [
                    f"SELECT {str(self)}",
                    "," + group_by_columns if group_by_columns != "" else "",
                    from_caluse,
                    group_by_clause,
                ]
            ),
            db=self._db,
            parents=parents,
        )
```

Does it make sense to treat them differently? Is there any concern to put the `order by` clause in the `query` str directly?

xuebinsu (approved):

greenplumpython/table.py
```diff
@@ -208,3 +231,16 @@ def values(rows: Iterable[Tuple], db: db.Database, column_names: Iterable[str] =
     rows_string = ",".join(["(" + ",".join(str(datum) for datum in row) + ")" for row in rows])
     columns_string = f"({','.join(column_names)})" if any(column_names) else ""
     return Table(f"SELECT * FROM (VALUES {rows_string}) AS vals {columns_string}", db=db)
+
+
+def order_by_str(order_by: Iterable) -> str:
```
beeender, line 236: Wouldn't this be a private functions?

beeender (commented):
Looks good except one comment

greenplumpython/table.py
Quasimewdo: I can make it private, the thing is that it doesn't depend on table.

beeender: uh, ok, it is an independent function.

Since it is unlikely to be used out of the `Table` class, I suggest making it a private static function of `Table`.

beeender (approved):

PR #54, opened 2022-06-09
Support Unary Operator on Column

Quasimewdo:
- Make table in test_binaryExpr.py temporary
- support not(), abs(), pos(), neg()
- Add tests
- TODO :
  * Integrate unary operator in table.__getitem__
  * Probably need to change how to do unary_operator_column name alias

### DIFF
--- a/greenplumpython/expr.py
+++ b/greenplumpython/expr.py
@@ -39,16 +39,16 @@ class Expr:
         return BinaryExpr("!=", self, other)
 
     def __pos__(self):
-        return UnaryExpr("+", self)
+        return UnaryExpr("+", self, as_name='"+' + str(self) + '"')
 
     def __neg__(self):
-        return UnaryExpr("-", self)
+        return UnaryExpr("-", self, as_name='"-' + str(self) + '"')
 
     def __abs__(self):
-        return UnaryExpr("ABS", self)
+        return UnaryExpr("ABS", self, as_name='"Abs(' + str(self) + ')"')
 
     def __invert__(self):
-        return UnaryExpr("NOT", self)
+        return UnaryExpr("NOT", self, as_name='"Not(' + str(self) + ')"')
 
     def __str__(self) -> str:
         raise NotImplementedError()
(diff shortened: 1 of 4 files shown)


greenplumpython/expr.py
```diff
@@ -64,6 +76,23 @@ def __str__(self) -> str:
...
+        super().__init__(as_name=as_name)
+        self.operator = operator
+        self.right = right
+
+    def __str__(self) -> str:
+        if self.operator == "NOT":
+            return "NOT(" + str(self.right) + ') AS "Not(' + str(self.right) + ')"'
+        if self.operator == "ABS":
+            return "ABS(" + str(self.right) + ') AS "Abs(' + str(self.right) + ')"'
+
```
beeender, line 90: we need to raise an exception for the unsupported operators. And, don't forget to add a test to check the exception.

greenplumpython/expr.py
```diff
@@ -64,6 +76,23 @@ def __str__(self) -> str:
         return str(self.left) + " " + self.operator + " " + str(self.right)
 
 
+class UnaryExpr(Expr):
+    def __init__(self, operator: str, right: Expr, as_name: Optional[str] = None):
+        super().__init__(as_name=as_name)
+        self.operator = operator
+        self.right = right
+
+    def __str__(self) -> str:
```
beeender, line 85: Why do you add `As "xxxx"` in this function? I think the `as_name` should be handled by the parent class `Expr`?

Quasimewdo: Changed it to use `as_name`

tests/test_unaryExpr.py
```diff
@@ -0,0 +1,61 @@
...
+    assert str(b3) == '-temp3.id AS "-temp3.id"'
+    ret = list(t[["id", str(b3)]].fetch())
+    for row in ret:
+        assert row["-temp3.id"] == -(row["id"])
+
+
+def test_expr_unary_abs(db: gp.Database):
+    rows = [(1,), (-2,), (-3,), (2,)]
+    t = gp.values(rows, db=db).save_as("temp4", column_names=["id"], temp=True)
+    b4 = abs(t["id"])
```
xuebinsu, line 48: How about
```python
b4 = abs(t["id"]).rename("abs")
```

xuebinsu (approved):

beeender (approved):
:clap:

PR #56, opened 2022-06-17
Support creating and calling array functions

xuebinsu:
Array functions take arrays as arguments. Before calling the
function, each attribute in the rows will be aggregated as an
array and passed to the function. In this way, the function can
have access to the attribute in all rows in the same call. This 
makes it possible for the user to do complex analytics and for
the PL handler to optimize the performance.


beeender (approved):
:clap:

Quasimewdo (approved):

PR #58, opened 2022-06-21
Support Return Composite Type for Func Call

Quasimewdo:
- Able to create composite type
- Able to call on composite type
- Able to return composite type

### DIFF
--- a/greenplumpython/table.py
+++ b/greenplumpython/table.py
@@ -1,4 +1,4 @@
-from typing import Iterable, List, Optional, Tuple
+from typing import Iterable, List, Optional, Tuple, Union
 from uuid import uuid4
 
 from . import db, expr
@@ -36,7 +36,18 @@ class Table:
         if isinstance(key, expr.Expr):
             return self.filter(key)
         if isinstance(key, slice):
-            raise NotImplementedError()
+            if key.step is not None:
+                raise NotImplementedError()
+            offset_clause = "" if key.start is None else f"OFFSET {key.start}"
+            limit_clause = (
+                ""
+                if key.stop is None
+                else f"LIMIT {key.stop if key.start is None else key.stop - key.start}"
+            )
+            return Table(
+                f"SELECT * FROM {self.name} {limit_clause} {offset_clause}",
+                parents=[self],
+            )
 
     def as_name(self, name_as: str) -> "Table":
         return Table(f"SELECT * FROM {self.name}", parents=[self], name=name_as, db=self._db)
@@ -51,6 +62,31 @@ class Table:
             f"""
                 SELECT {','.join([str(target) for target in target_list])} 
                 FROM {self._name}
+            """,
+            parents=[self],
+        )
+
+    @staticmethod
+    def _order_by_str(order_by: Iterable) -> str:
+        order_by_clause = (
+            f"""
+                {','.join([' '.join([col, order]) for col, order in order_by.items()])}
+            """
+            if isinstance(order_by, dict)
+            else f"""
+                    {','.join([order_index for order_index in order_by])}
+                """
+        )
+        return order_by_clause
+
+    def top(self, count: int, order_by: Iterable, skip: int = 0):
+        order_by_clause = self._order_by_str(order_by)
+        return Table(
+            f"""
+                SELECT * FROM {self.name}
+                ORDER BY {order_by_clause}
+                LIMIT {count}
+                OFFSET {skip}
             """,
             parents=[self],
         )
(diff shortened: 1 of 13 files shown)


tests/test_type.py
```diff
@@ -0,0 +1,32 @@
...
+        type.create_type(Person, "Person", db)
+    assert 'type "person" already exists\n' in str(exc_info.value)
+
+
+def test_type_drop(db: gp.Database):
+    class Person:
+        _first_name: str
+        _last_name: str
+
+    type.create_type(Person, "Person", db)
```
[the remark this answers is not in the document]
Quasimewdo: I am also thinking about it, maybe let the user to create the return type if it doesn't exist.

tests/test_func.py
```diff
@@ -201,3 +201,24 @@ def my_sum(val_list: List[int]) -> int:
...
+        ) -> None:
+            self._first_name = first_name
+            self._last_name = last_name
+
+    @gp.create_function
+    def create_person(first: str, last: str) -> Person:
+        return {"first_name": first, "last_name": last}
+
+    for row in create_person("'Amy'", "'An'", as_name="result", db=db).to_table().fetch():
+        print(row["result"])
```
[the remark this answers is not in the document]
Quasimewdo: Fixed

tests/test_func.py
```diff
@@ -201,3 +204,52 @@ def my_sum(val_list: List[int]) -> int:
...
+
+    @gp.create_function
+    def create_person(first: str, last: str) -> Person:
+        return {"first_name": first, "last_name": last}
+
+    for row in create_person("'Amy'", "'An'", db=db).to_table().fetch():
+        assert row["first_name"] == "Amy" and row["last_name"] == "An"
+
+
+def test_func_comp_type_column(db: gp.Database):
```
beeender, line 222: This test is not quite different from `test_func_return_comp_type`, what is the main purpose of this test? should it have a different name?
Is the original purpose of this test to test a composite-type column? but the column type is `int` I think?

tests/test_type.py
```diff
@@ -0,0 +1,82 @@
...
+    class Person:
+        _first_name: str
+        _last_name: str
+
+    class Couple:
+        _first_person: Person
+        _second_person: Person
+
+    def create_couple() -> Couple:
+        return Couple
```
beeender, line 74: it should return a `Couple` object instead of a `Couple` type.
Change it to `Couple()`

tests/test_type.py
```diff
@@ -0,0 +1,82 @@
...
+    class Person:
+        _first_name: str
+        _last_name: str
+
+    type.create_type(Person, db, as_name="Person", is_temp=False)
+    type.drop_type("Person", db)
+    type.create_type(Person, db, as_name="Person", is_temp=False)
+
+
+def test_create_type_rec(db: gp.Database):
```
beeender, line 64: does `rec` mean `recursive` here?

greenplumpython/table.py
```diff
@@ -1,6 +1,8 @@
 from typing import TYPE_CHECKING, Iterable, List, Optional, Tuple
 from uuid import uuid4
 
+from tabulate import tabulate
```
beeender, line 4: move to a separate PR.

tests/test_type.py
Quasimewdo: yes
```suggestion
def test_create_type_recursive(db: gp.Database):
```

Quasimewdo:
> There should be a type `Couple` in the signature, while an object `Couple()` in the return statement.

`    def create_couple() -> Couple:
        return Couple()`

greenplumpython/type.py
```diff
@@ -10,15 +15,50 @@
...
 
 
+# TODO : Add tests for all function
+def create_type(class_type, db, as_name=None, is_temp=True) -> str:
+    type_name = "type_" + uuid4().hex if as_name is None else as_name
+    temp_str = "pg_temp." if is_temp else ""
+    att_type_str = ",\n\t\t\t\t".join(
+        [
+            f"""
+            {re.sub(r"^_+|_+$", "", name)} {to_pg_type(type_t, db)}
```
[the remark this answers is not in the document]
Quasimewdo: There was an error when having underscore, but now it is ok
```suggestion
            {name} {to_pg_type(type_t, db)}
```

tests/test_func.py
```diff
@@ -51,16 +51,18 @@ def my_max(a: int, b: int) -> int:
...
-	def my_min(a: int, b: int) -> int:
-		if a < b:
-			return a
-		else:
-			return b
-
-	for row in my_min(1, 2, as_name="result", db=db).to_table().fetch():
-		assert row["result"] == min(1, 2)
-		assert row["result"] == inspect.unwrap(my_min)(1, 2)
+    @gp.create_function
```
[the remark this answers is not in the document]
beeender: not recovered :(

beeender (commented):
Please fix those minor issues then we are good to go !

PR #59, opened 2022-06-27
Sphinx-quickstart

Quasimewdo:
Documentation is essential for users to understand a project.
This patch adds the documentation for GreenpumPython's main class:
db, table and expr.


doc/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
Sasasu, line 1: windows support Orz....
I think only bash is enough

Sasasu (approved):

doc/make.bat
Quasimewdo: Will remove this file.

PR #60, opened 2022-06-30
Support passing Python objects and type casts

xuebinsu:
It is very common that an Expr contains Python objects. This patch
adds a wrapper to convert a Python object to a SQL constant
systematically. To ensure that the type of the const matches
what's desired, this patch also adds support for type casts by
"calling" instances of the Type class.

After adding the Type class, columns now can have types. This
opens the door to column type inference.

### DIFF
--- a/greenplumpython/func.py
+++ b/greenplumpython/func.py
@@ -8,7 +8,7 @@ from uuid import uuid4
 from .db import Database
 from .expr import Expr
 from .table import Table
-from .type import to_pg_const, to_pg_type, primitive_type_map
+from .type import primitive_type_map, to_pg_const, to_pg_type
 
 
 class FunctionCall(Expr):


greenplumpython/db.py
```diff
@@ -11,9 +11,9 @@ def __init__(self, **params) -> None:
             cursor_factory=psycopg2.extras.RealDictCursor,
         )
 
-    def execute(self, query: str, args: List = [], has_results: bool = True) -> Optional[Iterable]:
+    def execute(self, query: str, has_results: bool = True) -> Optional[Iterable]:
         with self._conn.cursor() as cursor:
-            cursor.execute(query, args)
```
Quasimewdo, line 16: We don't consider extra arguments anymore?

Quasimewdo (approved):

greenplumpython/expr.py
```diff
@@ -138,18 +132,48 @@ def __init__(
...
-        if self.operator == "NOT":
-            return "NOT(" + str(self.right) + ")"
-        if self.operator == "ABS":
-            return "ABS(" + str(self.right) + ")"
+        right_str = str(self.right) if isinstance(self.right, Expr) else to_pg_const(self.right)
+        return f"{self.operator}({right_str})"
 
-        return self.operator + str(self.right)
+
+class TypeCast(Expr):
```
Quasimewdo, line 139: Is this indentation correct for this class?

Quasimewdo: > It seems correct. The code is indented with 4 spaces.

Correct, my bad

PR #61, opened 2022-06-30
Support table display

Quasimewdo:
It is essential that user can read table's content clearly and quickly.
This patch supports table display using html format.


requirements.txt
```diff
@@ -1 +1,3 @@
 psycopg2==2.9.3
+
+tabulate
```
yihong0618: seems no need to add empty line.

And we need to fix `tabulate` version.

higuoxing:
It would be great if we can display a pretty table both in a Jupyter Notebook/Jupyter Lab environment and in an ordinary terminal. For users who don't use Jupyter Notebook, I think they also want a beautifully rendered table in ASCII. I would like to implement `__repr__` and `_repr_html_` for the `Table` class.

```python
class Table:
    ...
    def __repr__(self):
        return tabulate(self.fetch(), headers="keys", tablefmt="simple")

    def _repr_html_(self):
        return tabulate(self.fetch(), headers="keys", tablefmt="html")
```

So that we can get beautifully rendered table both in Jupyter environment and in an ordinary terminal.

[1] tabulate doc: https://pypi.org/project/tabulate/
[2] Rich display for IPython: https://ipython.readthedocs.io/en/stable/config/integrating.html?highlight=get_ipython#rich-display

higuoxing:
Why did you remove the 'tabulate' package? I think it can handle lots of uncovered situations (in our tests) when displaying tables.

yihong0618:
> Why did you remove the 'tabulate' package? I think it can handle lots of uncovered situations (in our tests) when displaying tables.

I prefer `[rich](https://github.com/Textualize/rich)`

Quasimewdo:
> Why did you remove the 'tabulate' package? I think it can handle lots of uncovered situations (in our tests) when displaying tables.

Because we want to implement our own `display` function without using external package

requirements.txt
```diff
@@ -1 +1 @@
-psycopg2==2.9.3
\ No newline at end of file
+psycopg2==2.9.3
```
beeender, line 1: what is changed?

Quasimewdo: I think nothing changed

higuoxing:
Just curious, what's the motivation of writing table display functions ourself? I think using either `rich` or `tabulate` is better.

greenplumpython/table.py
```diff
@@ -141,6 +140,34 @@ def __getitem__(self, *args, **kwargs):  # type: ignore
         """
         return self._getitem(*args, **kwargs)  # type: ignore
 
+    def __repr__(self):
+        """
+        Return a string representation for a table
+        """
+        repr_string = ""
+        ret = list(self.fetch())
+        repr_string += (("| {:10} |" * len(ret[0])).format(*ret[0])) + "\n"
```
higuoxing, line 149: What will happen when the length of the column names is greater than 10 charaters? It would be great to have some tests to cover the scenario. You don't need to do that in this PR but please try to improve it if you have ideas.

tests/test_const_table.py
```diff
@@ -60,3 +60,51 @@ def test_table_getitem_slice_offset(db: gp.Database, t: gp.Table):
...
+
+def test_table_display_repr(db: gp.Database):
+    # fmt: off
+    rows = [(1, "Lion",), (2, "Tiger",), (3, "Wolf",), (4, "Fox")]
+    # fmt: on
+    t = gp.values(rows, db=db).save_as("zoo1", column_names=["id", "animal"])
+    expected = (
+        "| id         || animal     |\n"
+        "============================\n"
+        "|          1 || Lion       |\n"
```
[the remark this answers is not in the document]
Quasimewdo: str is left aligned, and int is right aligned

greenplumpython/table.py
Quasimewdo: Got it, let's add a `FIXME` at the moment

higuoxing (approved):
+1

PR #62, opened 2022-07-07
Support Union two Table

Quasimewdo:
The UNION operator combines result sets of two or more `SELECT` 
statements into a single result set. This patch supports to union two 
tables into a new one, and with order by option.


greenplumpython/table.py
```diff
@@ -96,6 +96,32 @@ def top(self, count: int, order_by: Iterable, skip: int = 0):
             parents=[self],
         )
 
+    def union(
+        self,
+        other: "Table",
+        my_targets: List = [],
+        other_targets: List = [],
```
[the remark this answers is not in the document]
Quasimewdo: We can have some SQL like this:
`SELECT sex, UNNEST(col_nm) AS col_nm, UNNEST(coef) AS coef 
FROM plc_linreg_fitted
UNION 
SELECT sex, 'intercept' AS col_nm, intercept AS coef 
FROM plc_linreg_fitted;`
and to do so, with greenplumpython
`plc_linreg_fitted.union(
                plc_linreg_fitted,
                my_targets = [
                        plc_linreg_fitted["sex"],
                        unnest(plc_linreg_fitted["col_nm"]).rename("col_nm"),
                        unnest(plc_linreg_fitted["coef"]).rename("coef"),
                ],
                other_targets = [
                        plc_linreg_fitted["sex"],
                        "'intercept' AS col_name",
                        plc_linreg_fitted["intercept"].rename("coef"),
                ]
        )`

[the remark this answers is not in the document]
Quasimewdo: So we remove targets list for both table from union function?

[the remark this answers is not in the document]
Quasimewdo: Well, I think either we keep both, so the user can customize all with union function because it can be considered as one command when executing SQL command. Or we remove both for better reading, but users must to give extra names to save them.

Or maybe we can keep both scenarios, because with current version, we can also break the complex statement to several simple statements. 
```
plc_linreg_fitted_1 = plc_linreg_fitted[[
                        plc_linreg_fitted["sex"],
                        unnest(plc_linreg_fitted["col_nm"]).rename("col_nm"),
                        unnest(plc_linreg_fitted["coef"]).rename("coef")]
                ]
plc_linreg_fitted_2 = plc_linreg_fitted[[
                        plc_linreg_fitted["sex"],
                        "'intercept' AS col_name",
                        plc_linreg_fitted["intercept"].rename("coef")]
                ]
plc_linreg_fitted_1.union(plc_linreg_fitted_2)
```
This works

greenplumpython/table.py
```diff
@@ -96,6 +96,22 @@ def top(self, count: int, order_by: Iterable, skip: int = 0):
             parents=[self],
         )
 
+    def union(
+        self,
+        other: "Table",
+        all: bool = False,
```
yihong0618, line 102: is_all maybe better?
all somehow is a function

tests/test_union.py
```diff
@@ -0,0 +1,56 @@
...
+    assert cpt == 2
+
+
+def test_union_select(db: gp.Database, top_rated_films: gp.Table, most_popular_films: gp.Table):
+    top_rated_films = top_rated_films[
+        [top_rated_films["title"].rename("titltle"), top_rated_films["release_year"]]
+    ]
+    ret = list(top_rated_films.union(most_popular_films).fetch())
+    assert len(ret) == 5
+    assert list(list(ret)[0].keys()) == ["titltle", "release_year"]
```
yihong0618, line 56: seems we can change list(list(ret)[0].keys()) == ["titltle", "release_year"] -> list(ret[0].keys()) == ["titltle", "release_year"]
we list the `ret` before

greenplumpython/table.py
Quasimewdo: Changed `is_all` to `all` to keep it consistent with the signature of `fetch()`, maybe change both to `is_all` will be better? @xuebinsu

tests/test_union.py
Quasimewdo: > seems we can change list(list(ret)[0].keys()) == ["titltle", "release_year"] -> list(ret[0].keys()) == ["titltle", "release_year"] we list the `ret` before

That's right! Thanks.

yihong0618 (approved):
LGTM cool

PR #66, opened 2022-07-11
Fix CTEs Lineage error

Quasimewdo:


greenplumpython/table.py
```diff
@@ -200,17 +200,22 @@ def _list_lineage(self) -> List["Table"]:
...
-            for table_ in lineage[current]._parents:
-                if table_.name not in tables_visited and not table_._in_catalog():
-                    lineage.append(table_)
-                    tables_visited.add(table_.name)
+            if lineage[current].name not in tables_visited and not lineage[current]._in_catalog():
+                self._topologicalSort(lineage[current], tables_visited, lineage)
             current += 1
         return lineage
 
+    def _topologicalSort(self, t, visited, lineage):
```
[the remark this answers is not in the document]
Quasimewdo: Got it, changed it.

greenplumpython/table.py
```diff
@@ -200,17 +200,22 @@ def _list_lineage(self) -> List["Table"]:
         tables_visited = set()
         current = 0
         while current < len(lineage):
-            for table_ in lineage[current]._parents:
-                if table_.name not in tables_visited and not table_._in_catalog():
-                    lineage.append(table_)
-                    tables_visited.add(table_.name)
+            if lineage[current].name not in tables_visited and not lineage[current]._in_catalog():
```
beeender, line 203: The PR title is "Fix CTEs Lineage error". May I ask what the "Lineage error" was? I tried to use the code in the left and run the newly added two tests, they can still pass. I am a bit confused.

Quasimewdo: The order of CTE's aggregation was not correctly sorted when the operation on both tables for join becomes complex, we might have the below error
```
DETAIL:  There is a WITH item named "cte_f671f66477a44fd3b1cd494324926268", but it cannot be referenced from this part of the query.
HINT:  Use WITH RECURSIVE, or re-order the WITH items to remove forward references.
```
Do you mean with the old code, the two newly added tests can still pass? Maybe I need to add an extra operation on the last case.

beeender: > Do you mean with the old code, the two newly added tests can still pass? Maybe I need to add an extra operation on the last case.

Yes! With the old code, the two cases can still pass. So I got confused.

beeender (approved):
:clap: 

Please also put the original error message in the final commit message.

PR #67, opened 2022-07-11
Remove order by for all utilities

Quasimewdo:
In order to have a stable public interface, we decided to
remove support of ORDER BY statement for current
release.


beeender (approved):
You killed your own baby!

PR #68, opened 2022-07-12
Fix problem of Optional params when creating function

Quasimewdo:
Greenplum supports different language handler to create
python function. This patch supports to modify language handler 
and other optional parameters when creating functions.

### DIFF
--- a/greenplumpython/func.py
+++ b/greenplumpython/func.py
@@ -111,7 +111,7 @@ def aggregate(name: str, db: Database) -> Callable[..., FunctionCall]:
 
 # FIXME: Add test cases for optional parameters
 def create_function(
-    func: Callable,
+    func: Optional[Callable] = None,
     name: Optional[str] = None,
     schema: Optional[str] = None,
     temp: bool = True,
@@ -120,6 +120,18 @@ def create_function(
     return_type_as_name: Optional[str] = None,
     type_is_temp: bool = True,
 ) -> Callable:
+    if not func:
+        return functools.partial(
+            create_function,
+            name=name,
+            schema=schema,
+            temp=temp,
+            replace_if_exists=replace_if_exists,
+            language_handler=language_handler,
+            return_type_as_name=return_type_as_name,
+            type_is_temp=type_is_temp,
+        )
+
     @functools.wraps(func)
     def make_function_call(
         *args: Expr,
@@ -174,13 +186,23 @@ def create_function(
 
 # FIXME: Add test cases for optional parameters
 def create_aggregate(
-    trans_func: Callable,
+    trans_func: Optional[Callable] = None,
     name: Optional[str] = None,
     schema: Optional[str] = None,
     temp: bool = True,
     replace_if_exists: bool = False,
     language_handler: str = "plpython3u",
 ) -> Callable:
+    if not trans_func:
+        return functools.partial(
+            create_aggregate,
+            name=name,
+            schema=schema,
+            temp=temp,
+            replace_if_exists=replace_if_exists,
+            language_handler=language_handler,
+        )
+
     @functools.wraps(trans_func)
     def make_function_call(
         *args: Expr,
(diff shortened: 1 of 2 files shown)


greenplumpython/func.py
```diff
@@ -120,6 +120,18 @@ def create_function(
     return_type_as_name: Optional[str] = None,
     type_is_temp: bool = True,
 ) -> Callable:
+    if not func:
+        return functools.partial(
```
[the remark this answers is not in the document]
Quasimewdo:
```python
    assert sorted([row["result"] for row in results]) == [
        inspect.unwrap(multiply)(i, i) for i in range(10)
    ]
```
I can add this to line 110 in `test_func.py`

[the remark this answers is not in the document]
Quasimewdo: It seems like it still works. Can this test case cover it?
```python
def test_create_func_optional_params(db: gp.Database):
    @gp.create_function
    def multiply(a: int, b: int) -> int:
        return a * b

    @gp.create_function(name="multiply", replace_if_exists=True)
    def multiply2(a: int, b: int) -> int:
        return a * b * 2

    rows = [(i, i) for i in range(10)]
    series = gp.values(rows, db=db, column_names=["a", "b"])
    results = multiply2(series["a"], series["b"]).to_table().fetch()
    assert sorted([row["multiply"] for row in results]) == [i * i * 2 for i in range(10)]
    assert sorted([row["multiply"] for row in results]) == [
        inspect.unwrap(multiply2)(i, i) for i in range(10)
    ]
```

tests/test_func.py
```diff
@@ -94,6 +94,42 @@ def test_func_on_more_than_one_table(db: gp.Database):
     assert "Cannot pass arguments from more than one tables" == str(exc_info.value)
 
 
+def test_create_func_optional_params(db: gp.Database):
+    @gp.create_function
+    def multiply(a: int, b: int) -> int:
+        return a * b
+
+    @gp.create_function(replace_if_exists=True)
```
beeender, line 102: Is this test correct? I tried to change the `replace_if_exists` to `False`, and the test can still pass.

tests/test_func.py
```diff
@@ -94,6 +94,42 @@ def test_func_on_more_than_one_table(db: gp.Database):
...
+        inspect.unwrap(multiply)(i, i) for i in range(10)
+    ]
+
+
+def test_create_func_optional_params_name(db: gp.Database):
+    @gp.create_function
+    def multiply(a: int, b: int) -> int:
+        return a * b
+
+    @gp.create_function(name="multiply", replace_if_exists=True)
```
beeender, line 120: ditto

beeender (commented):
>  This patch supports to modify language handler

Better update the PR message when the commit changes. It took me a while to figure out it doesn't change anything with the language handler.

tests/test_func.py
Quasimewdo: Corrected

greenplumpython/func.py
```diff
@@ -120,6 +120,18 @@ def create_function(
     return_type_as_name: Optional[str] = None,
     type_is_temp: bool = True,
 ) -> Callable:
+    if not func:
```
beeender, line 123: To be honest, I have spent quite a while to understand why the `func` is `None` here, but did find an answer. Code like this deserves an inline comment.

Quasimewdo: Got it

tests/test_func.py
```diff
@@ -203,6 +274,21 @@ def my_sum(val_list: List[int]) -> int:
         assert ("is_even" in row) and (row["is_even"] is not None) and (row["result"] == 5)
 
 
+def test_array_func(db: gp.Database):
```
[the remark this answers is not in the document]
Quasimewdo: Got it

greenplumpython/func.py
beeender: uh, OK! I thought the `None` `func` means the same decorator is applied to the same name functions. :bulb:

PR #69, opened 2022-07-13
Fix Simple Pylance typing error

Quasimewdo:


beeender (commented):
Can the same warning be reported by pyright? It's better we have the linter set for CI so a similar issue can be caught in the future.

To change the configuration for pyright, change https://github.com/greenplum-db/GreenplumPython/blob/master/pyproject.toml

And here are some configurations for pyright: https://github.com/microsoft/pyright/blob/main/docs/configuration.md

Otherwise, maybe we can set up pylance on CI as well. Let me know if you need help on this.

Quasimewdo:
> Can the same warning be reported by pyright? It's better we have the linter set for CI so a similar issue can be caught in the future.
> 
> To change the configuration for pyright, change https://github.com/greenplum-db/GreenplumPython/blob/master/pyproject.toml
> 
> And here are some configurations for pyright: https://github.com/microsoft/pyright/blob/main/docs/configuration.md
> 
> Otherwise, maybe we can set up pylance on CI as well. Let me know if you need help on this.

We can add this `typeCheckingMode = "strict"` in `pyproject.toml`

beeender:
> We can add this typeCheckingMode = "strict" in pyproject.toml

I tried that at the beginning, but that was "too" strict at that time. Will it pass now with strict mode?

beeender:
> Well not yet, there are still lots of reports
That might be too much ...
If possible, we shall only enable the configs we care

beeender:
```
# pyproject.toml
[tool.pyright]
include = [
    "greenplumpython",
    "setup.py"
]
strict =[
    "greenplumpython",
]
```

Will cause more lint failures than what we fixed in this PR. Maybe we can enable some of the strict rules but not all of them. 

@Quasimewdo Let me know if you need help on this.

Quasimewdo:
I have added `typeCheckingMode = "strict"` in `pyproject.toml`, and set some report o `false`. 
I have also ignored some current unsolvable type error report by adding `#type : ignore` in code.
@beeender @xuebinsu

beeender (commented):
LGTM!
Need python experts on this @xuebinsu @yihong0618 

I am THE PY IDIOT

greenplumpython/util.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def get(self, fn, annotations, *args):
+        """get returns the matching function from the virtual namespace.
+        return None if it did not fund any matching function.
+        """
+        func = Function(fn)
+        return self.function_map.get(func.key(args=args, annotations=annotations))
+
+
+def overload(fn):
```
Quasimewdo, line 91: This can't overpass `Pyright` error

.github/workflows/lint.yml
```diff
@@ -25,6 +25,10 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Set up Node # for pyright
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12'
```
yihong0618, line 31: need to change with the pyright version

yihong0618 (approved):

PR #71, opened 2022-07-21
Add `group_by()` as a method of `Table`

xuebinsu:
This patch adds the `group_by()` method in the `Table` class, which
corresponds to the `GROUP BY` clause in SQL. Adding this method
helps further split a complex aggregation query into several simple
method calls, therefore making it easier for the user to express
such type of tasks.

### DIFF
--- a/doc/source/modules.rst
+++ b/doc/source/modules.rst
@@ -12,3 +12,5 @@ The **GreenplumPython** library contains 5 main modules:
    func
    op
    type
+   group
+   order


tests/test_group.py
```diff
@@ -0,0 +1,21 @@
...
+    yield db
+    db.close()
+
+
+def test_group_agg(db: gp.Database):
+    rows = [(i, i % 2 == 0) for i in range(10)]
+    numbers = gp.values(rows, db=db, column_names=["val", "is_even"])
+    count = gp.aggregate("count", db=db)
+
+    results = list(numbers.group_by("is_even").aggregate(count, numbers["val"]).to_table().fetch())
```
Quasimewdo, line 18: Why split `group_by` and `aggregate` if eventually we need to call `aggregate` function after `group_by`?

greenplumpython/group.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class TableRowGroup:
+    def __init__(self, table: "Table", group_by: Iterable["Expr"]) -> None:
+        self._table = table
+        self._group_by = [self._table[e] if isinstance(e, str) else e for e in group_by]
+
+    def aggregate(self, func: Callable, *args: "Expr") -> "FunctionExpr":
+        qualified_args = [self._table[e] if isinstance(e, str) else e for e in args]
+        return func(*qualified_args, group_by=self._group_by)
```
beeender, line 16: As a syntax sugar, this doesn't look too bad IMO.

My concern is , sooner or later, we need to define the built-in aggregates/functions like:

```
# gp/aggregates/defs.py
def count():
    return gp.aggregate("count")
```

so probably we need a proxy method here like:

```
class TableRowGroup:
    def count():
        return aggregate(gp.aggregates.count);
```

Since to create the `count` aggregate, a `db` obj needs to be passed as an argument. Then how do we define those built-in functions/aggregates?

beeender: > count() is already available in every db and we don't need to create it.

I know

> In fact, every "built-in" SQL functions that we want to create a Python method for should already be available in every db. That's what "built-in" means.

Exactly, how do we ship a `count()` method by default? How does the `def count()` look like in our code base?

tests/test_group.py
Quasimewdo: I think separate `group by` will certainly be easier to use.

tests/test_group.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@pytest.fixture
+def db():
+    db = gp.database(host="localhost", dbname="gpadmin")
+    yield db
+    db.close()
+
+
+def test_group_agg(db: gp.Database):
```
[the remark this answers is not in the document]
beeender: Does this mean there is going to be more changes on this PR? If so, consider make it as draft?

tests/test_group.py
```diff
@@ -0,0 +1,63 @@
...
+            and (row["is_even"] is not None)
+            and ("is_multiple_of_3" in row)
+            and (row["is_multiple_of_3"] is not None)
+        )
+        assert (
+            (row["is_even"] and row["is_multiple_of_3"] and row["count"] == 1)  # 0
+            or (row["is_even"] and not row["is_multiple_of_3"] and row["count"] == 2)  # 2, 4
+            or (not row["is_even"] and row["is_multiple_of_3"] and row["count"] == 1)  # 3
+            or (not row["is_even"] and not row["is_multiple_of_3"] and row["count"] == 2)  # 1, 5
+        )
```
higuoxing, line 40: I think we can just compare `row["is_even"]` and `row["is_multiple_of_3"]` with boolean values to make sure they are not `None`?
```suggestion
        assert (
            (row["is_even"] == True and row["is_multiple_of_3"] == True and row["count"] == 1)  # 0
            or (row["is_even"] == True and row["is_multiple_of_3"] == False and row["count"] == 2)  # 2, 4
            or (row["is_even"] == False and row["is_multiple_of_3"] == True and row["count"] == 1)  # 3
            or (row["is_even"] == False and row["is_multiple_of_3"] == False and row["count"] == 2)  # 1, 5
        )
```

higuoxing (commented):
Some of my comments, please correct me if they're not applicable.

greenplumpython/group.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class TableRowGroup:
+    def __init__(self, table: "Table", grouping_sets: List[List["Expr"]]) -> None:
+        self._table = table
+        self._grouping_sets = grouping_sets
+
+    # TODO: provide apply() instead of aggregate() for consistency. That is:
+    # `t.group_by(["is_even"])[["val"]].apply(count)`.
+    # But how about `t[[*]].apply(count)` ?
```
Quasimewdo, line 16: How to determine which column to llok for group with `*`?

greenplumpython/func.py
```diff
@@ -7,21 +7,22 @@
...
         func_name: str,
         args: Iterable[Expr] = [],
-        group_by: Optional[Iterable[Union[Expr, str]]] = None,
+        group_by: Optional[TableRowGroup] = None,
         as_name: Optional[str] = None,
         db: Optional[Database] = None,
         is_return_comp: bool = False,
     ) -> None:
-        table: Optional[Table] = None
+        table: Optional[Table] = group_by.table if group_by is not None else None
```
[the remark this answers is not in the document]
higuoxing: Oh, I didn't relize that we are using type checker. Please let it as is. I will come back later.

greenplumpython/group.py
```diff
@@ -0,0 +1,37 @@
...
+    from .func import FunctionExpr
+    from .table import Table
+
+
+class TableRowGroup:
+    def __init__(self, table: "Table", grouping_sets: List[List["Expr"]]) -> None:
+        self._table = table
+        self._grouping_sets = grouping_sets
+
+    def apply(self, func: Callable[["Table"], "FunctionExpr"]) -> "FunctionExpr":
```
Quasimewdo, line 14: Do we need to add a docstring for these functions?

Quasimewdo (approved):

higuoxing (approved):
+1

yihong0618 (approved):
LGTM

yihong0618:
is this need to add in `doc/source/groupby.rst` something like?

yihong0618:
@xuebinsu seems we need to follow new import way?

PR #77, opened 2022-08-05
Update README to include the pandas-like API

xuebinsu:
See the changed README.md for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,9 @@ GreenplumPython provides a [pandas](https://pandas.pydata.org/)-like table API t
 2. is powerful to do complex analytics, such as statistical analysis, with UDFs and UDAs
 3. encapsulates common best practices and avoids common pitfalls in Greenplum, compared to writing SQL directly
 
-## Installation
+## Getting Started
+
+### Installation
 
 To install the latest development version, do
 
@@ -19,9 +21,7 @@ NOTE: This version is considered UNSTABLE. DON'T use it in the production enviro
 
 Stable version will be released soon.
 
-## Tutorial
-
-### 1. Selecting the Database of Your Data
+### Selecting the Database of Your Data
 
 To begin with, we need to select the database that contains the data we want:
 
@@ -42,7 +42,7 @@ def display(t: gp.Table):
     return tabulate(t.fetch(), headers="keys", tablefmt="html")
 ```
 
-### 2. Accessing a Table in the Database
+### Accessing a Table in the Database
 
 After selecting the database, we can access a table in the database by specifying its name:
 
@@ -69,7 +69,7 @@ display(t)
 </tbody>
 </table>
 
-### 3. Basic Data Manipulation
+### Basic Data Manipulation
 
 Now we have a table. We can do basic data manipulation on it, just like in SQL.
 
@@ -142,7 +142,7 @@ Finally when we are done, we can save the resulting table to the database, eithe
 t_n.save_as(table_name="t_n", temp=True)
 ```
 
-### 4. `JOIN`-ing Two Tables
+### `JOIN`-ing Two Tables
 
 We can also `JOIN` two tables with GreenplumPython. For example, suppose we have two tables like this:
 
@@ -211,7 +211,7 @@ display(t_join)
 </tbody>
 </table>
 
-### 5. Creating and Calling Functions
+### Creating and Calling Functions
 
 Calling functions is essential for data analytics. GreenplumPython supports creating Greenplum UDFs and UDAs from Python functions and calling them in Python.
 


README.md
```diff
@@ -1,207 +1,304 @@
...
+```python
+from tabulate import tabulate
 
-### Description
 
- gpapply allows you to run a R function with input data frame in GreenplumDB through PL/container or PL/Python language.
- The Python function will be parsed, then an UDF will be created in GreenplumDB schema for execution.
- The calculation will be done in parallel with computing resources of GPDB segment hosts.
+def display(t: gp.Table):
+    return tabulate(t.fetch(), headers="keys", tablefmt="html")
```
higuoxing, line 28: It might be good to mention we are using the `jupyter notebook` or `jupyter lab` environment?

README.md
```diff
@@ -1,207 +1,304 @@
...
-```GPdatabase```     |   gp.GPDatabase: The GreenplumDB connection instance 
-```output_meta```     |    gp.GPTableMeta:  The output metadata (i.e. the details of output table). The output_meta.signature must be set. Please see below for the details of gp.GPTableMeta. If the output_meta.name is `None`, function will return a pd.dataframe as the results instead of store the results in a new table with name output_mate.name.
-```clear_existing```     |      Bool: Whether clear existing output table stored in GreenplumDB before executing the query 
-```runtime_id```     |      Used by "plcontainer" runtime type only. The runtime id is set by plcontainer to specify a runtime cnofiguration. See plcontainer for more information. The argument type is string, e.g. "plc_python_shared" 
-```runtime_type```     |      value should be "plcontainer" or "plpythonu" 
-```**kwargs```     |       if the function py_func contains extra arguments other than input table columns, you can append them after all of the required arguments list.
+Finally when we am done, we can save the resulting table to the database, either temporarily or persistently:
```
higuoxing, line 125:
```suggestion
Finally when we are done, we can save the resulting table to the database, either temporarily or persistently:
```

README.md
```diff
@@ -1,207 +1,304 @@
...
-# connect to GPDB
-conn = connection.connect(dbIP, dbPort, dbname, username, password, ...)
+GreenplumPython is a Python library that enables the user to interact with Greenplum in a Pythonic way.
 
-# init GPDB Database instance
-GPDatabase_instance = gp.GPDatabase(conn)
+GreenplumPython provides a pandas-like table API that
+1. looks familiar and intuitive to Python users
+2. is powerful to do complex analytics, such as statistical analysis, with UDFs and UDAs
+3. encapsulates common best practices and avoids common pitfalls in Greeenplum, compared to writing SQL directly
```
Quasimewdo, line 8:
```suggestion
3. encapsulates common best practices and avoids common pitfalls in Greenplum, compared to writing SQL directly
```

README.md
```diff
@@ -1,207 +1,304 @@
...
+</tbody>
+</table>
 
-# close all connecions
-connection.close_all()
+# Creating and Calling Functions
 
-# close one connection, -1 for false
-conn_id = connection.close_connection(conn_id)
+Calling functions is essential for data analytics. GreeenplumPython supports creating Greenplum UDFs and UDAs from Python functions and calling them in Python.
```
higuoxing, line 202:
```suggestion
Calling functions is essential for data analytics. GreenplumPython supports creating Greenplum UDFs and UDAs from Python functions and calling them in Python.
```

README.md
```diff
@@ -1,207 +1,304 @@
...
 
+# Accessing a Table in the Database
 
-### Usage
+After selecting the database, we can access a table in the database by specifying its name:
 
 ```python
-gpApply(dataframe_wrapper, py_func, GPdatabase, output_meta,
-            clear_existing = True, runtime_id = "plc_python_shared", runtime_type = "plpythonu",  **kwargs)
+t = gp.table("demo", db=db)
```
Quasimewdo, line 36: We are not going to present the chaining method?
`t = db.get_table("demo")`

README.md
higuoxing: Yes, I mean the `display()` function cannot print the table under other environment, e.g., in a terminal? It's ok to leave it as is :-)

README.md
```diff
@@ -1,207 +1,304 @@
...
+<tbody>
+<tr><td style="text-align: right;"> 10</td><td style="text-align: right;"> 10</td></tr>
+<tr><td style="text-align: right;">  4</td><td style="text-align: right;">  4</td></tr>
+<tr><td style="text-align: right;">  8</td><td style="text-align: right;">  8</td></tr>
+<tr><td style="text-align: right;">  2</td><td style="text-align: right;">  2</td></tr>
+<tr><td style="text-align: right;">  6</td><td style="text-align: right;">  6</td></tr>
+</tbody>
+</table>
+
+For a quick glance, we can `SELECT` the first N rows of a table, like this:
```
Quasimewdo, line 107: Do we need to specify that is unordered?

beeender:
NICE! :tada:

README.md
```diff
@@ -1,207 +1,304 @@
...
+</tbody>
+</table>
 
-# connect to GPDB
-conn = gp.connection.connect(dbIP, dbPort, dbname, username, password, ...)
+We can `JOIN` the two table like this:
 
-conn1 = gp.connection.connect(dbIP, dbPort, dbname, username, password, ...)
+```python
+t_join = t1.join(
```
Quasimewdo, line 174: Since we have different functions for different types of JOIN, will it be better to specify one type here and to list all other possibilities?

Quasimewdo (approved):
LGTM

higuoxing (approved):

PR #81, opened 2022-08-15
fix: path problems, add doc GitHub Page Actions

yihong0618:


tox.ini
```diff
@@ -39,3 +39,11 @@ commands =
     black . --check --diff
     isort --profile black . -c --diff
     pyright
+
+[testenv:docs]
+basepython = python3.9
+deps = {[base]deps}
+skip_install = true
+commands =
+    pip install .
```
beeender, line 49: not needed

requirements-dev.txt
```diff
@@ -3,3 +3,6 @@ isort==5.10.1
 pytest==7.0.1
 pglast==3.10
 pyright==1.1.250
+psycopg2==2.9.3
```
beeender: not needed, it is in the `requirments.txt` already.

doc/README.md
```diff
@@ -1,2 +1,8 @@
-Require `Sphinx`  
-Use `make builder` where `builer` can be `HTML`, `Latex`, ... 
\ No newline at end of file
+## Require `Sphinx`
```
beeender, line 1: remove this `README.md` and add a section in the root's readme about how to use tox to generate doc.

And remove the makefile in this dir.

.github/workflows/doc.yml
```diff
@@ -0,0 +1,53 @@
...
+    - name: Install dependencies
+      # pip install . --> for doc generate
+      run:  |
+        python -m pip install --upgrade pip
+        pip install -r requirements-dev.txt
+        pip install .
+    - name: Make doc
+      run: |
+         sphinx-build -M html doc/source build 
+         mkdir -p public
```
beeender, line 47: line 25-39 can be simply changed to `tox -e docs` I think?

yihong0618: done

doc/README.md
yihong0618: done

beeender (approved):

greenplumpython/__init__.py
```diff
@@ -1,8 +1,8 @@
-from .db import Database, database
-from .expr import get_type
-from .func import create_aggregate  # type: ignore
-from .func import create_array_function  # type: ignore
-from .func import create_function  # type: ignore
-from .func import aggregate, function
-from .op import operator
-from .table import Table, table, values
+from greenplumpython.db import Database, database
```
Quasimewdo, line 1: I tried with this syntax, but had error, was it because of line 54 `html_static_path = ["_static"]` in conf.py?

yihong0618: we need pip install . to install it as moudle

Quasimewdo (commented):
Brilliant, thanks!

greenplumpython/__init__.py
Quasimewdo: Learnt a new thing

PR #82, opened 2022-08-16
Add type annotations for built-in functions

xuebinsu:
SQL has several built-in functions that are widely used. For example,
count(), avg(), and max(). For better user experience, we add the
corresponding Python functions for them with full type annotations
so that the use will get better suggestions from their IDEs.

### DIFF
--- a/greenplumpython/table.py
+++ b/greenplumpython/table.py
@@ -576,12 +576,12 @@ class Table:
         """
         Apply a function to the table
         """
-        # We need to support calling functions with constant args or even no 
-        # arg. For example: SELECT count(*) FROM t; In that case, the 
-        # arguments do not conain information on any table or any database. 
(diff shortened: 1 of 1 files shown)


greenplumpython/table.py
```diff
@@ -576,7 +576,7 @@ def apply(self, func: Callable[["Table"], "FunctionExpr"]) -> "FunctionExpr":
         """
         Apply a function to the table
         """
-        return func(self)
+        return func(self)(table=self)
```
Quasimewdo, line 579: Why add the second parenthesis?

yihong0618 (approved):
LGTM

greenplumpython/table.py
Quasimewdo: Got it, thanx

Quasimewdo (approved):
+1

PR #85, opened 2022-08-24
feat: prepare for PyPI upload

yihong0618:


setup.py
```diff
@@ -1,28 +1,20 @@
...
-
-with open("requirements.txt") as f:
-    requirements = f.read().splitlines()
-
-
 setuptools.setup(
-    name="greenplum-python",
-    version="1.0.0-beta1",
-    install_requires=requirements,
+    name="greenplum-python-fork",
```
beeender, line 4: `-fork`

setup.py
```diff
@@ -1,28 +1,20 @@
...
+    version="1.0.0-beta4",
```
beeender, line 5: can we find a way to read the `VERSION` file? Or we need a way to extract the version number from this file and to tag automatically by concourse...

setup.py
```diff
@@ -1,28 +1,20 @@
...
+    install_requires=["psycopg2==2.9.3"],
```
beeender, line 6: toxwill run `setup.py`, correct? So maybe we don't need `requirements.txt` anymore.

yihong0618: yes we can add __version__

yihong0618: > yes we can add  “__version__”

setup.py
yihong0618: yes we can add “version”

setup.py
yihong0618: I think we'd better keep both of it.
Because `Python Guys` maybe just want to run locally they prefer to run `pip3 install -r requirements.txt`

beeender (approved):

PR #86, opened 2022-08-24
Fix array_func_group_by problem

Quasimewdo:
This patch fixes the problem of missing group_by columns in results
when creating and calling an array_function with group_by.


Quasimewdo:
> Better to add some test cases to guarantee that the behavior is expected and will not be changed in the future.

Noted

greenplumpython/func.py
```diff
@@ -144,12 +153,57 @@ class ArrayFunctionExpr(FunctionExpr):
...
+        self._extra_args = extra_args
+
     def _serialize(self) -> str:
         args_string = (
-            ",".join([f"array_agg({str(arg)})" for arg in self._args]) if any(self._args) else ""
+            ",".join(
+                [
+                    (
+                        f"array_agg({str(self._args[i])})"  # type: ignore
+                        if not self._extra_args[i]
```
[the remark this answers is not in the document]
Quasimewdo: Done, pls check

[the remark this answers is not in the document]
Quasimewdo: Yes, changed it

greenplumpython/func.py
```diff
@@ -146,10 +155,42 @@ class ArrayFunctionExpr(FunctionExpr):
...
+                            else str(self._args[i])  # type: ignore
+                        )
+                        if isinstance(self._args[i], Expr)
+                        else to_pg_const(self._args[i])  # type: ignore
+                    )
+                    for i in range(len(self._args))
+                ]
+            )
+            if any(self._args)
+            else ""
```
yihong0618, line 179: for this string does it better to split it to some lines of code.

This one-line python code is kind of difficult to read.

tests/test_func.py
```diff
@@ -415,3 +436,84 @@ def label(num: int, type_or_type: str) -> str:
...
+    results = list(
+        numbers.group_by("is_even").apply(lambda tab: my_stat(tab["val"])).to_table().fetch()
+    )
+    assert list(list(results)[0].keys()) == ["sum", "count", "is_even"]
+    for row in results:
+        assert (
+            ("is_even" in row)
+            and (row["is_even"] is not None)
+            and (row["sum"] == 5)
+            and (row["count"] == 5)
```
yihong0618, line 488: assert all(["is_even" in row, (row["is_even"] is not None, (row["sum"] == 5, (row["count"] == 5]) 
maybe easily to read?

yihong0618 (approved):
LGTM 
Only two comments maybe better for code read

tests/test_func.py
Quasimewdo: Changed it

greenplumpython/func.py
Quasimewdo: Got it, split it

PR #88, opened 2022-08-28
Support extend()-ing `Table` with extra column

xuebinsu:
This patch fixes several issues regarding `SELECT` that can cause
confusion or make the API inconsistent:

- `[[]]` now only accepts column names rather than arbitrary string,
  This helps to avoid SQL injection (but not eliminated yet).
- `select()` can still take args of arbitrary type and when the arg
  is a string, it means a string constant rather than a column name.
- Add `extend()` as a method of `Table` to include an extra value as
  a column. This helps reduce the need to refer to the columns of a
  table repeatedly.
- Remove the const wrapper `gp.rename()` as it can be covered by
  `extend()`

### DIFF
--- a/tests/test_join.py
+++ b/tests/test_join.py
@@ -313,9 +313,15 @@ def test_table_multiple_self_join(db: gp.Database, zoo_1: gp.Table):
         assert row["zoo2_animal"] == row["animal"]
 
 
-def test_join_recursive_join(db: gp.Database):
+# This test case is to guarantee that the CTEs are generated in the reversed
+# topological order (i.e. the DFS order) in the lineage graph.
+#
+# For that, we must ensure there are at least 4 nodes in the graph and the 2
+# direct parents of the final node must not be adjacent.
+def test_lineage_dfs_order(db: gp.Database):
     rows = [(i,) for i in range(10)]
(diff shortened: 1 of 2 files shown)


yihong0618 (approved):
Change LGTM.
only one question.
do we need expose all `serialize` function?

because there's still `_serialize` in class `ConstExpr`

tests/test_join.py
```diff
@@ -316,15 +316,6 @@ def test_table_multiple_self_join(db: gp.Database, zoo_1: gp.Table):
...
-            percentile,
-            (label["mod"] == percentile["mod"]),
-            targets=[label["val"], percentile["mod"]],
-        ).fetch()
-    )
-    assert len(ret_mod) == 5
-    for row in ret_mod:
-        assert row["mod"] == row["val"] % 2
+    label = numbers.extend("mod", numbers["val"] % 2)
+    results = label.inner_join(numbers, numbers["val"] == label["val"], targets=[label["val"]])
```
Quasimewdo, line 320: It seems like there has one layer less before join, this handle a specific recursive join test case.

Quasimewdo (approved):

PR #93, opened 2022-08-31
Modify advanced example adapting select and extend

Quasimewdo:
This patch set `session(autocommit=True)` and modifies DS advanced example of GreenplumPython:
- use `table.select()` and `table.extend()` to do select
- change way to cast rows to columns using `unnest(array_append))`
- remove `UNION` feature since it consider columns order rather than their name


beeender (approved):
:clap:

PR #95, opened 2022-09-02
Allow passing column names directly to `join()`

xuebinsu:
Preiously, `Table.join()` accepts generic `Expr`s as arguments. This
often requires the user to refer to the table again and again, which
can be annoying and error-prone.

This patch simplies join() by changing its arguments to column
names. The user can do renaming if necessary by passing dicts of
names. This makes join() less powerful but easier to use.

### DIFF
--- a/doc/source/notebooks/GreenplumPython-abalone.ipynb
+++ b/doc/source/notebooks/GreenplumPython-abalone.ipynb
@@ -52,6 +52,9 @@
    "metadata": {
     "pycharm": {
      "name": "#%%\n"
+    },
+    "vscode": {
+     "languageId": "sql"
     }
    },
    "outputs": [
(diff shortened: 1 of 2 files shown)


tests/test_join.py
```diff
@@ -11,134 +11,130 @@ def t1(db: gp.Database):
...
+    ret = next(
+        iter(
+            t1.join(
+                t2,
+                cond=t1["id1"] == t2["id2"],
+                self_columns=dict.fromkeys(["id1", "n1"]),
+                other_columns={"idd2"},
+            ).fetch()
+        )
+    )
```
yihong0618, line 86: there are quite some  next(iter(....)) can we make it to some property like 
 t1.join(
                t2,
                cond=t1["id1"] == t2["id2"],
                self_columns=dict.fromkeys(["id1", "n1"]),
                other_columns={"idd2"},
            ).fetch().row?

yihong0618 (approved):

beeender (approved):

PR #96, opened 2022-09-02
Fix error when displaying table with no results

Quasimewdo:
This patch fixes errors when `print(table)` for table's query has no results.


Quasimewdo (commented):
LGTM too!

PR #98, opened 2022-09-22
Move pr/merge pipeline to dev2

beeender:


concourse/pipeline/res_def.yml
```diff
@@ -107,21 +107,21 @@ resources:
 - name: bin_plpython3_gpdb6_rhel7
   type: gcs
   source:
-    bucket: gpdb-extensions-concourse-resources
-    json_key: ((extension/extensions-gcs-service-account-key))
-    versioned_file: intermediates/plpython3/plpython3_rhel7_gpdb6.gppkg
+    bucket: gp-extensions-ci
+    json_key: ((extension/extensions-gcs-service-account-key-dev2))
+    versioned_file: intermediates_release/plpython3/plpython3_rhel7_gpdb6.tar.gz
```
higuoxing, line 112: Seems that the pipeline of plpython3 puts versioned file to `intermediates/plpython3/plpython3_rhel7_gpdb6.tar.gz`?

beeender: yup, it packed a `tar.gz` file with version info in the name into another `tar.gz` file which doesn't have a version in the name, so we can use the versioned file. THe weird thing is totally because of the limitation of the concourse gcs resource.

higuoxing (approved):
:shipit:

zhrt123 (approved):

PR #99, opened 2022-09-27
Fix error when calling same UDF again

xuebinsu:
Previously, a UDF is not created until it is called. This would cause error when calling it more than once since the same function has already existed.

This patch fixes the error by introducing a global dict to keep track of which functions are defined and functions with the same name cannot be defined more than once.

Moreover, this patch removes support for creating non-temp functions and types. This helps make the results of Python scripts easier to reproduce by data scientists.

### DIFF
--- a/greenplumpython/func.py
+++ b/greenplumpython/func.py
@@ -467,7 +467,7 @@ def create_array_function(
     wrapped_func: Optional[Callable[..., Any]] = None, language_handler: str = "plpython3u"
 ) -> ArrayFunction:
     """
-    Creates a User Defined Array Function (UDF) in database from the given Python
+    Creates a User Defined Array Function in database from the given Python
     function.
 
     Args:


greenplumpython/func.py
```diff
@@ -170,292 +158,332 @@ def serialize(self) -> str:
...
+# It is not a Callable by design to prevent misuse.
+class _AbstractFunction:
+    def __init__(
+        self,
+        wrapped_func: Optional[Callable[..., Any]],
+        name: Optional[str],
+        schema: Optional[str],
+    ) -> None:
+        if not (
+            (name is not None and len(name) <= 63)
```
higuoxing, line 184: Maybe we can store 63 to a constant value? e.g., `NAMEDATALEN`?

greenplumpython/builtin/function.py
```diff
@@ -1,53 +1,53 @@
...
 
 
 def max(
     arg: Any,
     group_by: Optional[TableRowGroup] = None,
-    as_name: Optional[str] = None,
     db: Optional[Database] = None,
 ) -> FunctionExpr:
-    return FunctionExpr("max", [arg], group_by=group_by, as_name=as_name, db=db)
+    return FunctionExpr(aggregate(name="max"), (arg,), group_by=group_by, db=db)
```
higuoxing, line 31:
```suggestion
    return FunctionExpr(aggregate("max"), (arg,), group_by=group_by, db=db)
```

Can we omit the argument's name here?

greenplumpython/db.py
```diff
@@ -37,6 +40,7 @@ def execute(self, query: str, has_results: bool = True) -> Optional[Iterable[Tup
                 result = db.execute("SELECT version()")
 
         """
+        print(query)
```
higuoxing, line 43: Is it for debugging? Shall we remove it?
```suggestion
```

greenplumpython/func.py
```diff
@@ -170,292 +158,332 @@ def serialize(self) -> str:
...
+                    break
+        return db
 
-    """
 
-    def make_function_call(
-        *args: Expr, as_name: Optional[str] = None, db: Optional[Database] = None
-    ) -> FunctionExpr:
-        return FunctionExpr(name, args, as_name=as_name, db=db)
+_global_scope: Dict[str, _AbstractFunction] = {}
```
higuoxing, line 217: Can we put `_global_scope` at the top of this file, since it's a global variable?

greenplumpython/func.py
```diff
@@ -170,292 +158,332 @@ def serialize(self) -> str:
...
+            self._created_in_dbs.add(db)
+
+    def __call__(self, *args: Any, db: Optional[Database] = None) -> FunctionExpr:
+        if self._wrapped_func is not None:
+            _db = self._infer_db(args, db)
+            assert _db is not None, "Database is required to create function"
+            self._create_in_db(_db)
+        assert (
+            self._returning_composite is not None
+        ), f'Unknown if the return type of function "{self.qualified_name}" is composite.'
```
higuoxing, line 281:
```suggestion
        ), f'The return type of function "{self.qualified_name}" is expected to be composite.'
```

Maybe a better error message?

greenplumpython/func.py
```diff
@@ -170,292 +158,332 @@ def serialize(self) -> str:
...
+    def __call__(
+        self, *args: Any, group_by: Optional[TableRowGroup] = None, db: Optional[Database] = None
+    ) -> ArrayFunctionExpr:
+        if self._wrapped_func is not None:
+            _db = self._infer_db(args, db)
+            assert _db is not None, "Database is required to create array function"
+            self._create_in_db(_db)
+        assert (
+            self._returning_composite is not None
+        ), f'Unknown if the return type of array function "{self.qualified_name}" is composite.'
```
higuoxing, line 469: Ditto. Better error message?

greenplumpython/func.py
```diff
@@ -170,292 +158,332 @@ def serialize(self) -> str:
...
     Args:
-        func : python function
-        name : Optional[str] : name of the UDF
-        schema : Optional[str] : where the UDF will be stored
-        temp: bool : if the creation of the UDF is temporary exists only for current session
-        replace_if_exists : bool : if the creation replaces an existing UDF with same name and args
-        language_handler : str : language handler extension to create UDF in Greenplum, by default plpython3u, but can also choose plcontainer.
+        wrapped_func: python function
+        language_handler language handler to run the UDF, by default plpython3u,
+            but can also choose plcontainer.
```
higuoxing, line 486:
```suggestion
        language_handler language handler to run the UDF, defaults to plpython3u,
            but can also support plcontainer.
```

greenplumpython/func.py
```diff
@@ -170,292 +158,325 @@ def serialize(self) -> str:
...
+        """
+        Get the wrapped Python function in the database function.
+        """
+        assert self._wrapped_func is not None, "No Python function is wrapped inside."
+        return self._wrapped_func
+
+    def _create_in_db(self, db: Database) -> None:
+        assert self._created_in_dbs is not None
+        assert self._wrapped_func is not None
+        if db not in self._created_in_dbs:
```
beeender, line 244: Do we really want to handle multiple database cases (in the future)? from the implementation of `_try_resolve_db()`, it is not handled well.

greenplumpython/func.py
```diff
@@ -170,292 +158,325 @@ def serialize(self) -> str:
...
+            (name if schema is None else f"{schema}.{name}")
+            if name is not None
+            else f"pg_temp.{wrapped_func.__name__}"
+            if wrapped_func is not None
+            else None
+        )
+        assert qualified_name is not None
+        assert (
+            qualified_name not in _global_scope
+        ), f'Function named "{qualified_name}" has been defined before.'
```
beeender, line 199: nice check, just add a test case for this:

```
def test_create_func_same_name_throws(db: gp.Database):
    @gp.create_function
    def dup_name(a: int, b: int) -> int:
        return a + b

    with pytest.raises(Exception) as e:
        @gp.create_function
        def dup_name(a: int, b: int) -> int:
            return a + 1
    assert "has been defined before" in str(e.value)

```

greenplumpython/func.py
```diff
@@ -170,292 +158,325 @@ def serialize(self) -> str:
...
-            create_aggregate,  # type: ignore
-            name=name,
-            schema=schema,
-            temp=temp,
+    if transition_func is None:
+        return functools.partial(create_aggregate, language_handler=language_handler)
+    return AggregateFunction(
+        transition_func=NormalFunction(
+            transition_func,
+            name="func_" + uuid4().hex,
```
beeender, line 445: not for this PR scope, do we consider to use a shorter uuid? like 
https://github.com/skorokithakis/shortuuid

greenplumpython/func.py
```diff
@@ -170,292 +158,325 @@ def serialize(self) -> str:
...
-    temp: bool = True,
-    replace_if_exists: bool = False,
-    language_handler: str = "plpython3u",
-) -> Callable:  # type: ignore
+    wrapped_func: Optional[Callable[..., Any]] = None, language_handler: str = "plpython3u"
+) -> ArrayFunction:
     """
-    Creates a User Defined Python Function (UDF) in Greenplum Database according to the Python
-    function code. But it will array aggregate all args then pass them to the UDF.
+    Creates a User Defined Array Function (UDF) in database from the given Python
```
beeender, line 473: > User Defined Array Function (UDF)

Still UDF?

greenplumpython/func.py
```diff
@@ -170,292 +158,325 @@ def serialize(self) -> str:
...
+# The parent class for all database functions.
+# It is not a Callable by design to prevent misuse.
+class _AbstractFunction:
+    def __init__(
+        self,
+        wrapped_func: Optional[Callable[..., Any]],
+        name: Optional[str],
+        schema: Optional[str],
+    ) -> None:
+        NAMEDATALEN = 64  # See definition in PostgreSQL
```
yihong0618, line 183: NAMEDATALEN move to const not in this Class scope maybe better?

yihong0618 (commented):
amazing work!

greenplumpython/func.py
beeender: I think here is better, limiting the scope first and relaxing it when needed is always better IMO

higuoxing (approved):
LGTM! Please make sure other reviewers are happy with change too.

greenplumpython/func.py
beeender:
```
>>> shortuuid.uuid()
b'vytxeTZskVKR7C7WgdSP3d'

>>> base64.standard_b64encode(uuid4().bytes)
b'7tKLaUQoTR+PDeOszhDWCg=='

>>> uuid4().hex
b'a75f590faaa64d70b19a5f2cc9bf0778'
```

8 bytes shorter? Anyway, not an issue related to this PR

Quasimewdo (commented):
LGTM

greenplumpython/func.py
beeender: `_try_resolve_db()` always get the first `db` to use, what if the others are not the same with the first one?

PR #103, opened 2022-10-17
Add functional API

xuebinsu:
GreenplymPython's goal is to help user break a big SQL query into
several easy function calls in Python. However, this allows the user
to contruct incorrect queries. For example, the user can write an
`Expr` composed of columns from `Table`s from more than one
database. But such `Expr` cannot be evaluated by the server.

This patch tries to mitigate this issue by adding a functional API.
This makes it possible to delay binding `Expr`s and other objects
to `Table`s and `Database`s until the we know how to do it
automatically for the user. This also makes the API easier to
compose, and the entire program concise and fun to write.

### DIFF
--- a/greenplumpython/func.py
+++ b/greenplumpython/func.py
@@ -79,7 +79,7 @@ class FunctionExpr(Expr):
         it is an Aggregation function.
         """
         self.function.create_in_db(self._db)
-        from_caluse = f"FROM {self.table.name}" if self.table is not None else ""
+        from_clause = f"FROM {self.table.name}" if self.table is not None else ""
         group_by_clause = self._group_by.clause() if self._group_by is not None else ""
         parents = [self.table] if self.table is not None else []
         assert (
@@ -102,7 +102,7 @@ class FunctionExpr(Expr):
                 [
                     f"SELECT {str(self)}",
                     ("," + ",".join(grouping_cols)) if grouping_cols is not None else "",
-                    from_caluse,
+                    from_clause,
                     group_by_clause,
                 ]
             ),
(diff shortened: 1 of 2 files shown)


greenplumpython/func.py
```diff
@@ -75,20 +78,30 @@ def to_table(self) -> Table:
...
+        )
         orig_func_table = Table(
             " ".join(
                 [
                     f"SELECT {str(self)}",
-                    ("," + ",".join([str(target) for target in self._group_by.get_targets()]))
-                    if self._group_by is not None
-                    else "",
+                    ("," + ",".join(grouping_cols)) if grouping_cols is not None else "",
                     from_caluse,
```
Quasimewdo, line 105: from_clause

greenplumpython/table.py
```diff
@@ -201,21 +202,22 @@ def rename(self, name: str) -> "Table":
...
 
         Returns:
             Table : Table filtered according **expr** passed in argument
         """
-        return Table(f"SELECT * FROM {self._name} WHERE {str(expr)}", parents=[self])
+        return Table(f"SELECT * FROM {self._name} WHERE {str(predicate(self))}", parents=[self])
 
-    # FIXME: Add test
-    def select(self, *targets: Iterable[Any]) -> "Table":
+    def _select(self, targets: Callable[["Table"], Union[Any, Iterable[Any]]]) -> "Table":
```
Quasimewdo, line 217: Why only change `select()` to private but not also `where()`

Quasimewdo (approved):

greenplumpython/func.py
```diff
@@ -319,10 +342,12 @@ def transition_function(self) -> NormalFunction:
         ), f'Transition function of the aggregate function "{self.qualified_name}" is unknown.'
         return self._transition_func
 
-    def _create_in_db(self, db: Database) -> None:
+    def create_in_db(self, db: Database) -> None:
+        if self._transition_func is None:  # Function has already existed.
```
yihong0618, line 346: why we not use self.transition_function here since it is a property.
and a little confuse about the `comment` here.

yihong0618 (approved):

greenplumpython/func.py
yihong0618: copy that, thanks.

PR #107, opened 2022-10-21
Refactor how `Expr`s are related to `Table`

xuebinsu:
Previously, there are multiple ways Exprs are related to a Table, 
such as by assigning an Expr to a new column, or by `Expr.to_table()`
explicitly, etc. This makes the interface complicated and hard to
learn.

This patch solves the issue by unifying all these ways to using
`assign()` and removing all the others. With `assign()`, the user
can bind any Expr to a new column with name specified explicitly.
This makes the API simple and clear, and also making it convenient
for us to implement table column inference later.

### DIFF
--- a/tests/test_func.py
+++ b/tests/test_func.py
@@ -97,8 +97,8 @@ def test_simple_agg(db: gp.Database):
     numbers = gp.to_table(rows, db=db, column_names=["val"])
     count = gp.aggregate_function("count")
 
-    results = list(numbers.group_by().assign(count=lambda t: count(t["val"])))
-    assert len(results) == 1 and results[0]["count"] == 10
+    results = numbers.group_by().assign(count=lambda t: count(t["val"]))
+    assert len(list(results)) == 1 and next(iter(results))["count"] == 10
 
 
 def test_agg_group_by(db: gp.Database):
@@ -239,7 +239,7 @@ def test_array_func_group_by_return_composite(db: gp.Database):
         .assign(_sum=lambda t: t["result"]["_sum"], _count=lambda t: t["result"]["_count"])
     )
     assert all(e in next(iter(ret)).column_names() for e in ["_sum", "_count", "lab"])
-    for row in list(ret):
+    for row in ret:
         assert row["_sum"] == 3
         assert row["_count"] == 3
 
@@ -357,7 +357,7 @@ def test_func_apply_join(db: gp.Database):
         t2, cond=lambda t1, t2: t1["id1"] == t2["id2"], self_columns={"id1"}, other_columns={"n2"}
     )
     result = ret.assign(label=lambda t: label(t["n2"], t["id1"]))
-    for row in list(result):
+    for row in result:
         assert row["label"][1] == row["label"][2]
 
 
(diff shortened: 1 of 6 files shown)


greenplumpython/expr.py
```diff
@@ -24,13 +21,11 @@ class Expr:
...
     def __init__(
         self,
-        as_name: Optional[str] = None,
         table: Optional["Table"] = None,
         db: Optional[Database] = None,
     ) -> None:
-        self._as_name = as_name
         self._table = table
-        self._db = table.db if table is not None else db
+        self._db = db if db is not None else (table.db if table is not None else None)
```
yihong0618, line 28: self._db = db or table.db?

greenplumpython/group.py
```diff
@@ -20,22 +20,33 @@ def __init__(self, table: "Table", grouping_sets: List[Iterable["Expr"]]) -> Non
...
-                numbers.group_by("is_even").apply(lambda row: count(row["*"]))
+            :class:`Table` with the new columns.
         """
-        return func(self._table).bind(group_by=self)
+        from greenplumpython.table import Table
+
+        if len(new_columns) == 0:
+            return self
+        targets: List[str] = []
+        if len(new_columns):
```
yihong0618, line 39: we if len(new_columns) in the first and return early there is no need if len(new_columns) IMO.

greenplumpython/db.py
```diff
@@ -79,8 +79,21 @@ def table(self, name: str):
...
+        from greenplumpython.type import to_pg_const
+
+        targets: List[str] = []
+        for k, f in new_columns.items():
+            v: Any = f()
+            if isinstance(v, Expr):
+                assert v.table is None, "New column should not depend on any table."
+            if isinstance(v, FunctionExpr):
+                v = v.bind(db=self)
+            targets.append(f"{v.serialize() if isinstance(v, Expr) else to_pg_const(v)} AS {k}")
```
Quasimewdo, line 95: So we are supposing the result column name is compulsory when calling `assign()` like `assign(result_col=lambda)`

greenplumpython/table.py
```diff
@@ -249,34 +246,17 @@ def assign(self, **new_columns: Dict[str, Callable[["Table"], Any]]) -> "Table":
...
-            )
-        return Table(f"SELECT *, {','.join(new_columns_str)} FROM {self.name}", parents=[self])
+        targets: List[str] = []
+        if len(new_columns):
+            for k, f in new_columns.items():
+                v: Any = f(self)
+                if isinstance(v, Expr) and not (v.table is None or v.table == self):
+                    raise Exception("Newly included columns must be based on the current table")
+                targets.append(f"{v.serialize() if isinstance(v, Expr) else to_pg_const(v)} AS {k}")
+            return Table(f"SELECT *, {','.join(targets)} FROM {self.name}", parents=[self])
```
Quasimewdo, line 259: What if the user just want the result column of function and none of others, is this `*` necessary?

tests/test_func.py
```diff
@@ -94,46 +94,34 @@ def test_func_on_more_than_one_table(db: gp.Database):
...
 
-    results = list(
-        count(
-            numbers["val"],
-            group_by=numbers.group_by("is_even", "is_multiple_of_3"),
-            db=db,
-        )
-        .to_table()
+    results = (
+        numbers.group_by("is_even", "is_multiple_of_3")
```
Quasimewdo, line 123: So for **group_by**, we will only allow the syntax `tab.group_by.assign()`, no more `tab.assign(lambda t: func(t, group_by=t.group_by())`? (I am totally ok with this modification)

greenplumpython/expr.py
yihong0618: return None.
image

yihong0618: get your point you are right.

Quasimewdo (approved):
LGTM

yihong0618 (approved):
LGTM

PR #108, opened 2022-10-26
Support DISTINCT

xuebinsu:
This patch adds support for DISTINCT, both SELECT DISTINCT and distinct-qualified aggregation like count(DISTINCT). Specifically, distinct() are added to both `Table` and `AggregateFunction` as methods, respectively.

This patch also fixes error when assign() is empty after group_by(), changing it to truly return a `Table`, as specified in the signature.

### DIFF
--- a/greenplumpython/func.py
+++ b/greenplumpython/func.py
@@ -300,7 +300,7 @@ class AggregateFunction(_AbstractFunction):
         """
         Apply the current aggregate function to only each distinct set of the
         arguments.
-        
+
         For example, `count.distinct(t['a'])` means applying the
         `count()` function to each distinct value of :class:`Column` `t['a']`.
 


greenplumpython/func.py
```diff
@@ -44,23 +46,27 @@ def __init__(
...
         self._args = args
         self._group_by = group_by
+        self._distinct = distinct
 
     def bind(
         self,
         group_by: Optional[TableGroupingSets] = None,
         table: Optional[Table] = None,
         db: Optional[Database] = None,
+        distinct: Optional[bool] = None,
```
yihong0618, line 56: do we need add test for .bind(..., distinct=True) or something

greenplumpython/func.py
```diff
@@ -5,6 +5,7 @@
 import inspect
 import re
 import textwrap
+from dis import dis
```
yihong0618, line 8: no need to add this IMO

Quasimewdo (approved):
LGTM

yihong0618 (approved):
LGTM

PR #109, opened 2022-10-26
Update docstring for next beta release

Quasimewdo:
Complete and clear documentation is necessary for the user to understand 
how a product works. This patch updates the documentation of GreenplumPython 
with the latest feature modifications.


greenplumpython/col.py
```diff
@@ -1,3 +1,6 @@
+"""
+This  module manage a Column and sub column
```
yihong0618, line 2: seems after `This` have two spaces?

yihong0618 (approved):
wonderful job!

greenplumpython/col.py
Quasimewdo: That's right, thx

PR #116, opened 2022-11-11
Add doc on comparison with pandas

xuebinsu:
To clarify some important concepts, this patch adds a notebook for
comparison with pandas, and add a more thorough intro to "Table".

This patch also fixes a bug in `order_by()` by adding the `LIMIT` 
clause in all cases so that it can take effect.


greenplumpython/table.py
```diff
@@ -1,25 +1,34 @@
...
+  | a consequence,
 
-Since the content is cached locally, it will become stale once the Table gets modified by someone else on the database system.
-Therefore, you might want to use refresh() to sync the latest update.
-**Note that refresh() is also expensive.**
+    - Retrieving them from the database system can be expensive, and
+    - They might be modified concurrently by other users of the database system.
 
+  | Due to the first consequence, once the data of a :class:`Table` is fetched from
+  | the database system, it will be cached locally for later use.
```
beeender: > it will be cached locally for later use.

I don't quite get this, what does "cached locally" mean here? Is this the iterator case we discussed before?

beeender: hmm, OK. We may want to change the statement when we support cursor. caching data locally from MPP DB may sound scary to users.

beeender (approved):

Quasimewdo (approved):

PR #117, opened 2022-11-16
Clarify Table object conception

Quasimewdo:
It is important for user to understand the concept of greenplumPython's Table object. This patch
renames three methods related to Table on trying to better define GreenplumPyton Table and to 
clarify the difference between greenplumpython Table and Table in Greenplum.
- gp.to_table(rows, db) --> db.make_table(rows=rows)  # Make a new `Table` with row-oriented data
- t.save_as("table_2") --> t.save_into("table_2") # Save the **data** into a new `Table`
- db.table("table_2) --> db.open_table("table_2") # Open a persisted `Table`
- Make Table documentation more clear


greenplumpython/table.py
```diff
@@ -630,17 +633,19 @@ def _fetch(self, is_all: bool = True) -> Iterable[Tuple[Any]]:
         result = self._db.execute(to_json_table._build_full_query())
         return result if result is not None else []
 
-    def save_as(self, table_name: str, temp: bool = False, column_names: List[str] = []) -> "Table":
+    def save_into(
+        self, table_name: str, temp: bool = False, column_names: List[str] = []
+    ) -> "Table":
         """
-        Save the table to database as a real Greenplum Table
+        Save the **data** into a new :class:`Table`
```
beeender, line 640: I actually don't want to hide the import details from users. This function DOES save (or "materialize" ) data into the database, and the information should be contained by the function description.

Do we still consider using `VTable` as the class name for GrenplumPython?

Some suggestion:

Save the **data** into the database, and return a new instance of  :class:`Table` which represents the newly saved data.

greenplumpython/table.py
```diff
@@ -630,17 +633,19 @@ def _fetch(self, is_all: bool = True) -> Iterable[Tuple[Any]]:
...
 
         Args:
             table_name : str
```
beeender, line 643: Need a short description for the `table_name` arg here.

greenplumpython/table.py
```diff
@@ -630,17 +633,19 @@ def _fetch(self, is_all: bool = True) -> Iterable[Tuple[Any]]:
...
             temp : bool : if table is temporary
```
beeender, line 644: This is what i am worried, which may confuse users. `if table is temporary` , what is the `table` here?

greenplumpython/db.py
```diff
@@ -66,15 +66,15 @@ def set_config(self, key: str, value: Any):
         assert isinstance(key, str)
         self.execute(f"SET {key} TO {value}", has_results=False)
 
-    def table(self, name: str):
+    def open_table(self, name: str):
         """
-        Returns a :class:`~table.Table` using table name and self
+        Open a persisted :class:`~table.Table`
```
beeender, line 71: What does "persisted :class:`~table.Table` " mean?? A GPDB table/view?

beeender (commented):
> gp.to_table(rows, db) --> db.make_table(rows=rows) # Make a new Table with row-oriented data
+1

greenplumpython/table.py
Quasimewdo: table_name: str: name of the table which can be later used with db.open_table(table_name) 
```suggestion
            table_name : str
```

greenplumpython/table.py
Quasimewdo: I think we can consider `VTable`.

Here is the description of verticapy's `vDataFrame.to_db`: 

> Saves the vDataFrame current relation to the Vertica Database. This method can also be used to materialize flex tables.

beeender:
Shall we use `UnifiedTable` as we discussed before? Or is there any progress on the naming?

greenplumpython/db.py
```diff
@@ -138,6 +138,32 @@ def assign(self, **new_columns: Callable[[], Any]) -> "Table":
...
+    def make_table(self, rows: Iterable[Tuple[Any]], column_names: Iterable[str] = []) -> "Table":
+        """
+        Makes a new :class:`Table` with row-oriented data
+
+        Args:
+            rows: Iterable[Tuple[Any]]: List of values
+            column_names: Iterable[str]: List of given column names
+
+        Returns:
+            Table: table made
```
yihong0618, line 150: return table is fine I think

yihong0618 (approved):

PR #119, opened 2022-11-25
Feat: set up to display sql or not

Quasimewdo:
To understand the result of GreenplumPython, users could
want to investigate the SQL query sent by GreenplumPython
to the database. This patch adds a global method to set up the option "sql_on".
If "sql_on" is True, GreenplumPython will print the query in the output, 
otherwise not. The default value is False.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -50,4 +50,5 @@ passenv=
 commands =
     # for doc generate we need to install this moudle that can `import greenplumpython`
     pip install --force-reinstall .
+    pip install --force-reinstall psycopg2-binary
     sphinx-build -W --keep-going -b html -j 2 {toxinidir}/doc/source doc/build/{env:TAG_REF:}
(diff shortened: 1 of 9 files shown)


Quasimewdo:
> Looks good overall. Please check why the CI failed.

Fixed

tox.ini
```diff
@@ -50,4 +50,5 @@ passenv=
 commands =
     # for doc generate we need to install this moudle that can `import greenplumpython`
     pip install --force-reinstall .
+    pip install --force-reinstall psycopg2-binary
```
yihong0618, line 53: can you add some comments why we need psycopg2-binary here

yihong0618 (approved):

tox.ini
Quasimewdo: Yes, I will explain it fix problem for macos when building docs in local evironment

PR #120, opened 2022-11-29
Support `in` operator

xuebinsu:
Membership testing is essential in data analytics. Previously, due to
lacking of the `in` operator, users have to use `join()` followed by
`distinct()` to get the correct result, which can be inefficient.

This patch adds the `in` operator, enabling using semi-join for
efficient membership testing, but with some limitations:
- Nested `in` exprs, e.g. `t["a"].in_(t2.["b"].in_(t3["c"]))` is not
supported. And only one table other than the current is allowed.
- `in` exprs can only be used to form operator exprs but no others
such as function calls.

### DIFF
--- a/greenplumpython/table.py
+++ b/greenplumpython/table.py
@@ -254,7 +254,6 @@ class Table:
         parents = [self]
         if v.other_table is not None and self.name != v.other_table.name:
             parents.append(v.other_table)
-        print("parents =", [t.name for t in parents])
         return Table(f"SELECT * FROM {self._name} WHERE {v.serialize()}", parents=parents)
 
     def apply(


greenplumpython/table.py
```diff
@@ -242,31 +249,13 @@ def where(self, predicate: Callable[["Table"], "Expr"]) -> "Table":
...
-                FROM {self._name}
-            """,
-            parents=[self],
-        )
+        v = predicate(self)
+        assert v.table == self, "Predicate must based on current table"
+        parents = [self]
+        if v.other_table is not None and self.name != v.other_table.name:
+            parents.append(v.other_table)
+        print("parents =", [t.name for t in parents])
```
Quasimewdo, line 257: Can we remove this `print`?

yihong0618 (approved):

PR #122, opened 2022-12-02
Fix Join default select empty problem

Quasimewdo:
This patch sets the default select columns of join to all. 
It removes also `Table.rename()` so the API can handle itself 
self join, the user doesn't need to handle rename Table.


beeender (approved):

Quasimewdo:
> Please update the commit message to reflect the latest change of this PR.

Done

PR #128, opened 2023-01-01
Fix repr for zeros + new pandas-like methods

ahmedrachid:
- Fix: representation of Zeros (0) by changing the "c" value condition in both _repr_ and repr_html from "if c" to "if c is not None" 
- Added pandas-like functions such as _describe_, _head_, _tail_, _size_, _shape_, _to_dataframe_ and _from_dataframe_


Quasimewdo:
I think you can update the name and description of this PR

PR #129, opened 2023-01-04
fix: zeros representation in table

ahmedrachid:
Noticed that Zeros (0) are not shown in the actual state because __repr__ and _repr_html_ are using "if c" condition instead of "if c is not None"


Quasimewdo (approved):
Hi, LGTM.
Maybe can add a testcase in test_const_table.py to handle this specific case (print table containing zero)

PR #131, opened 2023-01-09
Fix error when table in non-default schema

xuebinsu:
When retrieving data of a table in a non-default schema with `to_json()`,
an error will occur. This is because the SQL parser treats schema name
as table name and table name as column name and finds the table with
the schema name does not exist.

This patch fixes the error by renaming the table with the `AS` clause
to avoid ambiguity.

This patch also adds a test case for converting pandas DataFame to
GreenplumPython Table without introducing pandas as dependency.

### DIFF
--- a/greenplumpython/table.py
+++ b/greenplumpython/table.py
@@ -558,32 +558,24 @@ class Table:
     def __next__(self):
         """:meta private:"""
 
-        def detect_duplicate_keys(json_pairs: List[tuple[str, Any]]):
+        def tuple_to_dict(json_pairs: List[tuple[str, Any]]):
             key_count = collections.Counter(k for k, _ in json_pairs)
             duplicate_keys = ", ".join(k for k, v in key_count.items() if v > 1)
 
             if len(duplicate_keys) > 0:
                 raise Exception("Duplicate column_name(s) found: {}".format(duplicate_keys))
-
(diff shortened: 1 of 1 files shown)


Quasimewdo (approved):
LGTM

requirements-dev.txt
```diff
@@ -3,3 +3,4 @@ isort==5.10.1
 pytest==7.0.1
 pglast==3.10
 pyright==1.1.250
+pandas
```
yihong0618, line 6: pandas is a little big do we need make it optional?

tests/test_table.py
```diff
@@ -292,3 +292,18 @@ def test_table_distinct(db: gp.Database):
...
+    pg_class = db.table("pg_catalog.pg_class")
+    assert len(list(pg_class)) > 0
+
+
+import pandas as pd
+
+
+def test_table_to_pandas_dataframe(db: gp.Database):
+    pg_class = db.table("pg_catalog.pg_class")
+
```
yihong0618, line 307: import in the function maybe better?

beeender (approved):

requirements-dev.txt
yihong0618: got it

yihong0618 (approved):

PR #135, opened 2023-01-16
Refactor how DataFrame is formatted

xuebinsu:
This patch fixes type errors in the formatting method of a DataFrame. Those errors are previously ignored.

This patch also adds row numbers to the formatted string so that DataFrame can be formatted even when it is empty.

Moreover, this patch also makes Row of DataFrame looks more like immutable dict.

### DIFF
--- a/greenplumpython/dataframe.py
+++ b/greenplumpython/dataframe.py
@@ -1,27 +1,29 @@
 """
-`DataFrame` is the core data structure in GreenplumPython. Conceptually, a `DataFrame`
-is a two-dimensional unordered structure containing data.
+`DataFrame` is the core data structure in GreenplumPython.
 
-In the data science world, a `DataFrame` in GreenplumPython is similar to a `DataFrame` in `pandas
-<https://pandas.pydata.org/>`_, except that
+Conceptually, a `DataFrame` is a two-dimensional structure containing data.
 
-- | Data in a GreenplumPython `DataFrame` is lazily evaluated rather than eagerly. That is, they are computed only when
-  | they are observed. This can improve efficiency in many cases.
-- | Data in a GreenplumPython `DataFrame` is located and manipulated on a remote database system rather than locally. As
-  | a consequence,
+In the data science world, a :class:`DataFrame` in GreenplumPython is similar to a `DataFrame
+<https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html>`_ in `pandas <https://pandas.pydata.org/>`_,
+except that:
 
-    - | Retrieving them from the database system can be expensive. Therefore, once the data 
-      | of a GreenplumPython :class:`DataFrame` is fetched from the database system, it will be cached locally for later use.
-    - | They might be modified concurrently by other users of the database system. You might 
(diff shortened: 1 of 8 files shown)


greenplumpython/row.py
```diff
@@ -1,15 +1,17 @@
...
 This module creates a Python object :class:`Row` for GreenplumPython DataFrame iteration.
 """
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, Iterable, List, Tuple, Union
 
 
 class Row:
     """
-    Represents a row of GreenplumPython :class:`~table.DataFrame`.
+    Represents a row of GreenplumPython :class:`~dataframe.DataFrame`.
```
beeender, line 9: remove `GreenplumPython`

tests/test_db.py
```diff
@@ -29,8 +29,11 @@ def test_db_get_dataframe(db: gp.Database):
...
-    gp.config.print_sql = False
-    assert gp.config.print_sql is False
+    assert gp.config.print_sql is not None
+
+
+def test_db_assign_empty(db: gp.Database):
+    df = list(db.assign())
+    assert len(df) == 1
+    for row in df:
+        assert len(row) == 0
```
beeender, line 39: Shouldn't this show `(0 rows)\n`? I got blank output:

```
df = db.assign()
print(repr(df))
```

beeender: If it has zero column, how could it have `(1 row)`? Would it be 0 row as well?

tests/test_dataframe.py
```diff
@@ -145,30 +157,36 @@ def test_dataframe_display_repr_html(db: gp.Database):
...
+        "-----------------------\n"
         " id           | animal \n"
-        " ------------ + ------ \n"
+        "--------------+--------\n"
         " [1, 1, 1]    |        \n"
         " [2, 2, 2]    | Tiger  \n"
         " [3, 3, 3]    |        \n"
         " [4, None, 4] | Fox    \n"
+        "-----------------------\n"
         "(4 rows)\n"
```
beeender, line 187: Can we add a test to show `(0 rows)`? I failed since we cannot do that with `from_rows()`. (the empty `VALUES` problem).

beeender (approved):

beeender:
oops, conflicts!

greenplumpython/dataframe.py
```diff
@@ -180,68 +180,83 @@ def __getitem__(self, _):
...
+        # To align each column, we use a two-pass algorithm:
+        # 1. Iterate over the DataFrame to find the max width for each column; and
+        # 2. Convert the datum in each column to str within the width.
+        first_row: Row = contents[0]
+        widths = {col: len(col) for col in first_row} if len(first_row) > 0 else {None: 2}
+        for row in contents:
+            for name, val in row.items():
+                widths[name] = max(widths[name], len(str(val)))
+        # For Python >= 3.7, dict.items() and dict.values() will preserves the
+        # input order.
```
yihong0618, line 204: we do not support python3.7 here

yihong0618 (approved):
LGTM

PR #137, opened 2023-01-16
Lint for python docstr

beeender:
- Add tox target docstyle which uses pydocstyle to lint python doc
  strings. See https://github.com/PyCQA/pydocstyle and
  http://www.python.org/dev/peps/pep-0257/
  'tox -e docstyle' will check all py files in greenplumpython
  directory.
  Use 'tox -e docstyle -- <path_to_the_py_file>' to lint individual py
  source file.
- The doc linter has not been enabled on CI yet. It will be done in
  another PR later.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,4 +3,5 @@ isort==5.10.1
 pytest==7.0.1
 pglast==3.10
 pyright==1.1.250
+pydocstyle==6.2.3
 pandas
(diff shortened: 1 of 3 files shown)


requirements-dev.txt
```diff
@@ -3,4 +3,5 @@ isort==5.10.1
 pytest==7.0.1
 pglast==3.10
 pyright==1.1.250
+pydocstyle==6.2.3
```
yihong0618, line 6: why not in requirements-doc?

yihong0618 (approved):

PR #142, opened 2023-01-17
Support outside dependencies for UDFs

xuebinsu:
Previously, to create a UDF in database, the user is required to write
all the dependencies inside the function body. This often results in
annoying boilerplate code

This patch adds support for refering to objects outside of the UDF,
including functions and modules. This closes the gap between a UDF
and a normal Python function.

This is achieved by using a pickling library, dill, to serialize a Python
function as well as all of its dependencies before sending it to
server. However, it requires to install the library on server to
deserialize the function before executing it. If the library is not 
usable, we will fallback automatically to the "source code" solution
by also sending the source code of the function to the server.

### DIFF
--- a/greenplumpython/func.py
+++ b/greenplumpython/func.py
@@ -340,9 +340,10 @@ class NormalFunction(_AbstractFunction):
 
 def function(name: str, schema: Optional[str] = None) -> NormalFunction:
     """
-    A wrap in order to call function
+    A wrap in order to call function (Predefined in-Database UDF) on :class:`~db.Dataframe`.
 
     Example:
+        Get a wrapper of the in-Database function `generate_series`
         .. code-block::  Python
 
             generate_series = gp.function("generate_series")
@@ -432,7 +433,7 @@ class AggregateFunction(_AbstractFunction):
 
 def aggregate_function(name: str, schema: Optional[str] = None) -> AggregateFunction:
     """
-    A wrap in order to call an aggregate function
+    A wrap in order to call an aggregate function (Predefined in-Database UDA).
 
     Example:
         .. code-block::  Python
@@ -447,8 +448,7 @@ def create_function(
     wrapped_func: Optional[Callable[..., Any]] = None, language_handler: str = "plpython3u"
 ) -> NormalFunction:
     """
-    Creates a User Defined Function (UDF) in database from the given Python
-    function.
+    Creates a User Defined Function (UDF) in database from the given Python function.
 
     Args:
         wrapped_func : the Python function to be wrapped into a database function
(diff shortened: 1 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -15,18 +15,20 @@ env:
...
-    name: Test P${{ matrix.python-version }}, PSQL${{matrix.postgres-version}}
+    name: Test Python${{ matrix.python-version }}, PSQL${{matrix.postgres-version}}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10"]
-        postgres-version: [10, 11, 12]
+        python-version: ["3.9"]
+        postgres-version: ['12', '13']
```
yihong0618, line 24: postgres 10 11 use default python3 version is python3.7 that we do not suport

.github/workflows/test.yml
```diff
@@ -47,6 +49,14 @@ jobs:
...
         python-version: ${{ matrix.python-version }}
         cache: pip
+    - name: Install Dependencies for greenplumpython dill
+      # plpytyhon for postgres install to pg services
+      run: |
+          # docker ps -a
+          # docker exec -i $(docker ps -q) python3 -m pip install dill
+          pip install pip install --target=. dill
+          docker cp dill $(docker ps -q):/usr/lib/${{ matrix.dill_version }}/
+          rm -rf dill pip
```
yihong0618, line 59: its a trick way that we do not need to build new images

.github/workflows/test.yml
```diff
@@ -15,18 +15,20 @@ env:
...
   build:
-    name: Test P${{ matrix.python-version }}, PSQL${{matrix.postgres-version}}
+    name: Test Python${{ matrix.python-version }}, PSQL${{matrix.postgres-version}}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10"]
-        postgres-version: [10, 11, 12]
+        python-version: ["3.9"]
```
yihong0618, line 23: need same python version so for now we drop  python310

greenplumpython/func.py
```diff
@@ -261,41 +266,57 @@ def create_in_db(self, db: Database) -> None:
             return
         assert self._created_in_dbs is not None
         if db not in self._created_in_dbs:
+            func_src: str = dill.source.getsource(self._wrapped_func)
+            func_ast: ast.FunctionDef = ast.parse(dedent(func_src)).body[0]
+            # TODO: Lambda expressions are NOT supported since inspect.signature()
+            # does not work as expected.
+            assert isinstance(
+                func_ast, ast.FunctionDef
+            ), f"{self._wrapped_func} is not a function. (lambda is not supported.)"
```
yihong0618, line 275: is a lambda expressions that we do no support for now?

yihong0618 (approved):

yihong0618:
LGTM
do we need mention the version should the same when using `dill` in the doc

Quasimewdo (approved):
LGTM

PR #151, opened 2023-02-06
Document utils for functions and types

xuebinsu:
This patch documents the usage and notes when using the utils for
creating and using functions and types. It also renames some
possibly confuing concepts, including

- `as_name` to `column_name` to avoid the concept of "alias"
- `ArrayFunction` to `ColumnFunction` to reflect its use case

It also does other minor renaming and checks to improve
consistency of the interface.

### DIFF
--- a/tests/test_func.py
+++ b/tests/test_func.py
@@ -665,7 +665,7 @@ def test_agg_composite_type(db: gp.Database):
     numbers = db.create_dataframe(rows=rows, column_names=["val"])
     result = numbers.group_by().apply(lambda t: sum_count(t["val"]), expand=True)
     for row in result:
-        assert row["count"] == len(rows) and row["sum"] == 10
+        assert row["count"] == len(rows) and row["sum"] == 45
 
 
 import math


yihong0618 (approved):
LGTM

FYI seems we did not change the `change` in `doc/source/notebooks/basic.ipynb`
code in this doc seems still the old one

yihong0618:
LGTM~

greenplumpython/order.py
```diff
@@ -80,7 +79,7 @@ def __getitem__(self, rows: slice) -> "DataFrame":
         Example:
             .. code-block::  Python
 
-                t.order_by("id").head(5)
+                t.order_by("id")[:5]
```
Quasimewdo, line 82: I am thinking to open a PR to change all `t` in example to `df`

tests/test_func.py
```diff
@@ -657,7 +659,7 @@ class sum_count_type(SimpleNamespace):
     def sum_count(result: sum_count_type, val: int) -> sum_count_type:
         if result is None:
             return SimpleNamespace(sum=val, count=1)
-        return SimpleNamespace(sum=result["sum"] + val, count=result["count"] + 1)
+        return SimpleNamespace(sum=result["count"] + val, count=result["count"] + 1)
```
Quasimewdo, line 662: Why `sum=result["count"] + val`?

Quasimewdo (approved):
LGTM! Very informative

yihong0618 (approved):

PR #152, opened 2023-02-08
Replace `= any()` with `EXISTS` for `in_()`

xuebinsu:
Per the PostgreSQL guideline
https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_NOT_IN,
`NOT IN` and `!= any()` has some intricate issues and should 
not be used in general.

Therefore, this patch replace the `= any()` with `EXISTS` in
implementation to
- make the semantic more clear , and
- make it easier for DB to generate efficient Anti-Join plans.

### DIFF
--- a/greenplumpython/expr.py
+++ b/greenplumpython/expr.py
@@ -495,7 +495,7 @@ class InExpr(Expr):
         container_name: str = "cte_" + uuid4().hex
         return (
             (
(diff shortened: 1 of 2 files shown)


beeender (approved):

yihong0618 (approved):
LGTM

PR #154, opened 2023-02-13
Rename all private methods by adding leading _

Quasimewdo:
Rename all private methods by adding leading `_` so that they would not be called by user

### DIFF
--- a/greenplumpython/col.py
+++ b/greenplumpython/col.py
@@ -36,7 +36,7 @@ class ColumnField(Expr):
         return self._column
 
     def _serialize(self) -> str:
-        return f'({self.column._serialize()})."{self._field_name}"'  # type: ignore
+        return f'({self.column._serialize()})."{self._field_name}"'
 
 
 class Column(Expr):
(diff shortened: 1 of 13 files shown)


greenplumpython/group.py
```diff
@@ -234,7 +234,7 @@ def dataframe(self) -> "DataFrame":
         return self._dataframe
 
     # FIXME: Make this function package-private
```
beeender, line 236: FIXME like this can be removed.

greenplumpython/col.py
```diff
@@ -35,8 +35,8 @@ def __init__(
     def column(self) -> "Column":
         return self._column
 
-    def serialize(self) -> str:
-        return f'({self.column.serialize()})."{self._field_name}"'
+    def _serialize(self) -> str:
+        return f'({self.column._serialize()})."{self._field_name}"'  # type: ignore
```
beeender, line 39: Since `reportPrivateUsage = false` has been set, the line suppression is not needed I think?

beeender (approved):
With on minor comment. Turning off `reportPrivateUsage` annoys me, but it seems to better way to avoid that.

PR #164, opened 2023-02-16
Fix doc & examples in db.py

beeender:


greenplumpython/conftest.py
```diff
@@ -32,6 +32,11 @@ def init_namepsace(doctest_namespace: Dict[str, Any]):
     doctest_namespace["gp"] = gp
     doctest_namespace["pd"] = pd
     doctest_namespace["cursor"] = cursor
+    doctest_namespace["db_host"] = _DBHOST
+    doctest_namespace["db_port"] = _DBPORT
+    doctest_namespace["db_name"] = _DBNAME
+    doctest_namespace["db_user"] = _DBUSER
+    doctest_namespace["db_password"] = _DBPSWD
```
yihong0618, line 39: how about make this to a config dict
doctest_namespace.update(config_dict)

yihong0618 (approved):

greenplumpython/conftest.py
beeender: oooooops ... merged ...

I think this is better since the example https://github.com/greenplum-db/GreenplumPython/blob/dea99ea0b14235f6710e0e49b06161cc0a73a716/greenplumpython/db.py#L250 here will be (a little bit) simpler

yihong0618: copy that.

PR #168, opened 2023-02-16
Fix doc & examples for func.py

beeender:
All future examples should be tested after this PR.
Also, some properties has been changed to private since it is not the time
to expose them yet.


greenplumpython/func.py
```diff
@@ -154,7 +156,7 @@ def apply(self, expand: bool = False, column_name: Optional[str] = None) -> Data
         )
 
     @property
-    def function(self) -> "_AbstractFunction":
+    def _function(self) -> "_AbstractFunction":
```
Quasimewdo, line 159: Since we add `_` before the name, maybe we can remove this property and use `self._func` directly

greenplumpython/func.py
```diff
@@ -494,10 +529,11 @@ def aggregate_function(name: str, schema: Optional[str] = None) -> AggregateFunc
         and :code:`schema`.
 
     Example:
+        .. highlight:: python
         .. code-block::  Python
 
             >>> count = gp.aggregate_function("count")
-            >>> db.create_dataframe(columns={'x': range(10)}).apply(lambda t: count())
+            >>> db.create_dataframe( columns={'x': range(10)}).group_by().apply(lambda t: count())
```
Quasimewdo, line 536:
```suggestion
            >>> db.create_dataframe(columns={'x': range(10)}).group_by().apply(lambda t: count())
```

greenplumpython/func.py
```diff
@@ -494,10 +529,11 @@ def aggregate_function(name: str, schema: Optional[str] = None) -> AggregateFunc
         and :code:`schema`.
 
     Example:
+        .. highlight:: python
         .. code-block::  Python
 
             >>> count = gp.aggregate_function("count")
```
Quasimewdo, line 535:
```suggestion
            >>> import greenplumpython.builtins.functions as F
```

greenplumpython/func.py
```diff
@@ -494,10 +529,11 @@ def aggregate_function(name: str, schema: Optional[str] = None) -> AggregateFunc
         and :code:`schema`.
 
     Example:
+        .. highlight:: python
         .. code-block::  Python
 
-            >>> count = gp.aggregate_function("count")
-            >>> db.create_dataframe(columns={'x': range(10)}).apply(lambda t: count())
+            >>> import greenplumpython.builtins.functions as F
+            >>> db.create_dataframe(columns={'x': range(10)}).group_by().apply(lambda t: F.count())
```
beeender, line 536: actually, this is not the right example anymore.

Quasimewdo (approved):

PR #169, opened 2023-02-16
fix: doc and examples in order.py

yihong0618:


greenplumpython/order.py
```diff
@@ -1,6 +1,4 @@
-"""
-This module creates a Python object DataFrameOrdering for order by dataframe.
-"""
+"""This module creates a Python object DataFrameOrdering for ordering by dataframe."""
```
beeender, line 1: Just remove this, it doesn't mean too much.
```
# noqa: D100
```

greenplumpython/order.py
```diff
@@ -12,7 +10,8 @@
 
 class DataFrameOrdering:
     """
-    Represents an ordered GreenplumPython :class:`~dataframe.DataFrame` generated by
+    Representation of an ordered GreenplumPython :class:`~dataframe.DataFrame` generated by.
```
beeender, line 13: Line 13 and line 15 was ONE sentence before, they cannot be splited.
Change to:

```
Represent an ordered GreenplumPython :class:`~dataframe.DataFrame`.

The order is decided by:
```

greenplumpython/order.py
```diff
@@ -68,18 +68,28 @@ def order_by(
 
     def __getitem__(self, rows: slice) -> "DataFrame":
         """
-        Returns a GreenplumPython :class:`~dataframe.DataFrame` that contains the slice of DataFrame in order.
+        Select parts of the :class:`~dataframe.DataFrame` ordering.
```
beeender, line 71:
```
Returns a slice of :class:`~dataframe.DataFrame` in the specified order.
```

greenplumpython/order.py
```diff
@@ -101,14 +111,15 @@ def __getitem__(self, rows: slice) -> "DataFrame":
     @property
     def dataframe(self) -> "DataFrame":
         """
-        Returns GreenplumPython :class:`~dataframe.DataFrame` associated for ORDER BY
+        Dataframe property.
 
         Returns:
-            GreenplumPython DataFrame
+            :class:`~dataframe.DataFrame` associated with ORDER BY
```
beeender, line 117: This is not the ordered dataframe, it is the original one.
Actually I think we can delete this property, just use `self._dataframe`.

@xuebinsu @Quasimewdo

Quasimewdo: Agree

greenplumpython/order.py
```diff
@@ -12,7 +10,9 @@
 
 class DataFrameOrdering:
     """
-    Represents an ordered GreenplumPython :class:`~dataframe.DataFrame` generated by
+    Represent an ordered GreenplumPython :class:`~dataframe.DataFrame`.
+
+    The order is decided by:
```
beeender, line 15: image

the generated doc seems confusing.

beeender (approved):

PR #171, opened 2023-02-20
Remove redundant attributes and property

Quasimewdo:
This patch removes redundant private `@property`. 
And fix type error of `_qualified_name`'s return type.
This patch removes also redundant attributes in class constructors
and methods.

### DIFF
--- a/greenplumpython/expr.py
+++ b/greenplumpython/expr.py
@@ -21,7 +21,7 @@ class Expr:
         # noqa: D107
         self._dataframe = dataframe
         self._other_dataframe = other_dataframe
-        self._db = db if db is not None else (dataframe.db if dataframe is not None else None)
+        self._db = db if db is not None else (dataframe._db if dataframe is not None else None)
 
     def __hash__(self) -> int:
         # noqa: D105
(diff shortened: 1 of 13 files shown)


greenplumpython/expr.py
```diff
@@ -553,7 +527,7 @@ def in_(self, container: Union["Expr", List[Any]]) -> "InExpr":
                 --------------
                 (5 rows)
         """
-        return InExpr(self, container, self.dataframe, self.db)
+        return InExpr(self, container, self._dataframe, self._db)
```
yihong0618, line 530: this self._db can drop too?

greenplumpython/func.py
```diff
@@ -61,14 +60,15 @@ def bind(
...
-        return FunctionExpr(
+        f = FunctionExpr(
             self._func,
             self._args,
             group_by=group_by,
             dataframe=dataframe,
-            db=db if db is not None else self._db,
             distinct=self._distinct,
         )
+        f._db = db if db is not None else self._db
```
yihong0618, line 70: assert self._db is not None here?

yihong0618 (commented):
amazing work!

greenplumpython/func.py
Quasimewdo: `assert db is not None or self._db is not None`?

yihong0618: yes.

Quasimewdo: Done

yihong0618 (approved):

yihong0618:
LGTM

PR #172, opened 2023-02-21
fix: drop db=db

yihong0618:
drop db=db in expr

for now we can not easily drop `db=db` in function expr, because we have decorator create_in_db use it.
and clean some imports


yihong0618:
Same work as #171 closing it now

PR #183, opened 2023-03-07
Set psycopg2-binary version requirement not strict

Quasimewdo:
Previously in `requirements.txt`, `psycopg2-binary==2.9.5`. 
This patch removes the strict version requirement and changes 
it to `2.9.*` to meet more possibility of versions.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-psycopg2-binary==2.9.5
+psycopg2-binary==2.9.*
 # Pin the setuptools version. The latest one seems to have problem with CFLAGS.
 # See https://github.com/pypa/setuptools/commit/bd7613f7921e8d60fa089d3ab419d0f04db6db6f
 setuptools==65.5.1


yihong0618 (approved):

beeender (changes requested):
Actually using the specific dependency version is on purpose -- to make sure the CI gives us a consistent result at a specific SHA1. 

If there is a need to bump the dependency version, just make a commit to bump it. I would like to avoid wildcard in the requirement.txt.

Does `2.9.5` have any problem?

Quasimewdo:
> Actually using the specific dependency version is on purpose -- to make sure the CI gives us a consistent result at a specific SHA1.
> 
> If there is a need to bump the dependency version, just make a commit to bump it. I would like to avoid wildcard in the requirement.txt.
> 
> Does `2.9.5` have any problem?

Well there is ask to build the package with `conda`, but the latest version of `psycopg2`-binary in `conda` is 2.9.3, so can't build it.

beeender:
> > Actually using the specific dependency version is on purpose -- to make sure the CI gives us a consistent result at a specific SHA1.
> > If there is a need to bump the dependency version, just make a commit to bump it. I would like to avoid wildcard in the requirement.txt.
> > Does `2.9.5` have any problem?
> 
> Well there is ask to build the package with `conda`, but the latest version of `psycopg2`-binary in `conda` is 2.9.3, so can't build it.

I was wrong, the `requirements.txt` is more about how user install it, but not what will be released.

This is OK.

beeender (approved):

PR #185, opened 2023-03-09
Fix Ambiguous table name in different schema

Quasimewdo:
This patch solves problems below:
- ambiguous table name in different schema
- table in non-default schema for advanced scenarios

### DIFF
--- a/greenplumpython/func.py
+++ b/greenplumpython/func.py
@@ -104,9 +104,7 @@ class FunctionExpr(Expr):
             expand and column_name is not None
         ), "Cannot assign single column name when expanding multi-valued results."
         self._function._create_in_db(self._db)
-        from_clause = (
-            f"FROM {self._dataframe._qualified_name_str}" if self._dataframe is not None else ""
-        )
+        from_clause = f"FROM {self._dataframe._name}" if self._dataframe is not None else ""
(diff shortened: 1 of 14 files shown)


tests/test_type.py
```diff
@@ -7,15 +7,28 @@
 from tests import db
 
 
+# Since we set all table_name to cte_hex(), we can't use regclass(t["tableoid"]) anymore
```
Quasimewdo, line 10: Does Greenplum have lots of operators or functions which will access to `tableoid`?

yihong0618 (approved):

greenplumpython/func.py
```diff
@@ -90,9 +90,7 @@ def _serialize(self) -> str:
             if any(self._args)
             else ""
         )
-        schema, func_name = self._function._qualified_name
-        qualified_name = f'"{schema}"."{func_name}"' if schema is not None else f'"{func_name}"'
-        return f"{qualified_name}({distinct} {args_string})"
+        return f"{self._function._qualified_name_str}({distinct} {args_string})"
```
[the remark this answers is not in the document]
Quasimewdo: Remaining `_qualified_name_str` are for `function` and `type` name

tests/test_type.py
Quasimewdo: I will just remove this comments

PR #186, opened 2023-03-10
Make Row a subclass of abc.Mapping

xuebinsu:
Conceptually, a Row is an immutable `dict` in Python and we previously added many methods to make this claim true. However, pandas requires stating its base class explicitly. Otherwise it will fail to treat it as an immutable `dict`.

To fix this issue, this patch sets the base class of Row to abc.Mapping explicitly and add other methods required by abc.Mapping


yihong0618 (approved):

Quasimewdo (approved):
LGTM

PR #187, opened 2023-03-15
Support GPDB 7

xuebinsu:
This patch creates jobs to our PR pipeline to verify that
GreenplumPython works flawlessly with GPDB 7.

Note that on GPDB 7, we are using the python3 provided by the
OS vender and we don't need to install python3 ourselves.

### DIFF
--- a/concourse/pipeline/release.yml
+++ b/concourse/pipeline/release.yml
@@ -2,10 +2,11 @@
 #@   "entrance_job",
 #@   "build_test_job",
 #@   "exit_release_job",
-#@   "centos6_gpdb6_conf",
 #@   "centos7_gpdb6_conf",
 #@   "rhel8_gpdb6_conf",
-#@   "ubuntu18_gpdb6_conf")
+#@   "ubuntu18_gpdb6_conf",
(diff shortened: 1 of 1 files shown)


concourse/pipeline/job_def.lib.yml
```diff
@@ -29,6 +29,14 @@ res_plpython3_bin: bin_plpython3_gpdb6_ubuntu18
 os: ubuntu1804
 #@ end
 
+#! Job config for gpdb7 rocky8
```
beeender, line 32: since we have both gpdb6 and gpdb7, please modify the `build_test_job_name` so the job name can show the gpdb version. It is a bit confusing otherwise.

concourse/pipeline/dev.yml
beeender, line 1: `pr.yml` and `commit.yml` need to be modified accordingly.

beeender: `release.yml` as well, yes. Although we don't have a release pipeline on concurse now.

beeender (approved):

higuoxing (approved):
LGTM!

PR #188, opened 2023-03-17
Fix display issue of README on PyPI

xuebinsu:
Previsouly, PyPI displayed the README as plain text. This is because it does not know the content type of the file.

This patch fixes the issue by adding a argument in setup() to tell PyPI to render the README as Markdown.

Reference: 
https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/


beeender (approved):
CI fails

beeender (approved):

yihong0618 (approved):

PR #189, opened 2023-03-17
Remove `--user` in install tuto

Quasimewdo:
Anaconda recommends do not use pip with the `--user` argument 
in `Conda` environment, avoid all “users” installs. 
To be more clear, this patch removes `--user` argument  in doc
and precise installation need to be with non root user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,16 +9,16 @@ GreenplumPython provides a [pandas](https://pandas.pydata.org/)-like DataFrame A
 
 ## Installation
 
-To install the latest development version, do
+To install the latest development version as non root user, do
 
 ```bash
-pip3 install --user git+https://github.com/greenplum-db/GreenplumPython
+pip3 install git+https://github.com/greenplum-db/GreenplumPython
 ```
 
-To install the latest released version, do
+To install the latest released version as non root user, do
 
 ```bash
-pip3 install --user greenplum-python
+pip3 install greenplum-python
 ```
 
 ## Documentation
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -9,16 +9,16 @@ GreenplumPython provides a [pandas](https://pandas.pydata.org/)-like DataFrame A
 
 ## Installation
 
-To install the latest development version, do
+To install the latest development version as non root user, do
```
beeender, line 12: Won't be this conflict with the below examples?

*as non root*, then install without `--user`. This would fail in the normal environment since normal user won't be able to write to the `/usr/local/`

doc/source/install.rst
```diff
@@ -5,26 +5,38 @@ GreenplumPython currently requires at least Python 3.9 to run, this is because:
...
 
-    pip3 install git+https://github.com/greenplum-db/GreenplumPython
+    pip3 install --user git+https://github.com/greenplum-db/GreenplumPython
 
-GreenplumPython requires `plpython3 <https://docs.vmware.com/en/VMware-Tanzu-Greenplum/6/greenplum-database/GUID-analytics-pl_python.html>`_ 
-extension to be installed on Greenplum/Postgres.
+With Virtual Environments
+-------------------------
+
+The `--user` option in an active virtual environment will install to the local user python location.
```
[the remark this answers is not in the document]
Quasimewdo: For example: `pip install greenplum-python`?

README.md
Quasimewdo: You're right, just added a new section for virtual env

beeender (approved):

yihong0618 (approved):

PR #194, opened 2023-05-25
Support caching in SD for internal use

xuebinsu:
In PL/Python, SD is used to cache objects between two function calls. However, when using dill with `recurse=True`, dill will overwrite `globals()` so that SD is removed.

This patch adds `SD` back as a member of `plpy` when using dill. After that, `SD` can be accessed with
```
sys.modules['plpy']._SD
```
However, IMHO, writing caching code manually for each UDF can be cumbersome. I don't think this interface is user-friendly enough to expose as public API.  Thus, this is marked as "for internal use" and no doc is added for it.


yihong0618 (approved):
LGTM

PR #197, opened 2023-06-14
Add distribution key for saving dataframe in db

ruxuez:
Greenplum has a MPP architecture, so it relies on distribution 
of data across segments. This patch supports defining distribution 
type and distribution key of table when saving Dataframe to GPDB.

### DIFF
--- a/greenplumpython/dataframe.py
+++ b/greenplumpython/dataframe.py
@@ -877,8 +877,8 @@ class DataFrame:
             storage_params: storage_parameter of gpdb, reference
                 https://docs.vmware.com/en/VMware-Tanzu-Greenplum/7/greenplum-database/GUID-ref_guide-sql_commands-CREATE_TABLE_AS.html
             schema: schema of the table for avoiding name conflicts.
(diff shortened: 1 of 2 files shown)


yihong0618 (approved):

greenplumpython/dataframe.py
```diff
@@ -919,12 +923,24 @@ def save_as(
             f"WITH ({','.join([f'{key}={storage_params[key]}' for key in storage_params.keys()])})"
         )
         df_full_name = f'"{table_name}"' if schema is None else f'"{schema}"."{table_name}"'
+        distribution_clause = (
+            f"""
+                DISTRIBUTED {f"BY ({','.join(distribution_key)})" 
+                if distribution_type == "hash" 
+                else "REPLICATED"
+                if distribution_type == "replicated"
+                else "RANDOMLY"}
```
yihong0618, line 932: do we have other distribution_type here?

ruxuez: We have hash, replicated, randomly and None

PR #203, opened 2023-06-27
Fix bind db

ruxuez:
This patch binds db from outside of call by adding `bind()`
in all components.


tests/test_type.py
```diff
@@ -33,6 +33,44 @@ def test_type_modifier(db: gp.Database):
...
+        lambda t: float8(func(t["a"], t["b"])),
+        column_name="float8",
+    )
+    assert sorted([row["float8"] for row in results_app]) == list(range(0, 20, 2))
+
+    results_ass = df.assign(float8=lambda t: float8(func(t["a"], t["b"])))
+    assert sorted([row["float8"] for row in results_ass]) == list(range(0, 20, 2))
+
+
+def test_type_cast_apply(db: gp.Database):
```
ruxuez, line 55: @xuebinsu Hi, just added a new testcase which applies a func on type cast result

yihong0618 (approved):

PR #207, opened 2023-07-25
Add module for indexing and searching embeddings

xuebinsu:
Vector embeddings are essential for searching unstructured data, such
as images and texts. Typically, to implement semantic search with
embeddings, user will need to go through a complex process:

1. Generate embeddings with transformers;
2. Save the generated embeddings into a table;
3. Create index on the generated embeddings;
4. Search for items with most similar embeddings.

To simplify the process, this patch adds a new module that provides
only two functions: `create_index()` and `search()`. This hides all
the embedding-related details and enables user to focus only on the
text or image data that they are really interested in.

### DIFF
--- a/greenplumpython/experimental/embedding.py
+++ b/greenplumpython/experimental/embedding.py
@@ -13,19 +13,19 @@ def _generate_embedding(content: str, model_name: str) -> gp.type_("vector"):  #
 
     import sentence_transformers  # type: ignore reportMissingImports
 
-    SD = globals().get("SD")
-    if SD is None:
-        SD = sys.modules["plpy"]._SD
-    if "model" not in SD:
+    sd_ = globals().get("SD")
+    if sd_ is None:
+        sd_ = sys.modules["plpy"]._SD
+    if "model" not in sd_:
         import torch  # pyright: ignore [reportMissingImports, reportUnknownVariableType]
 
         # Limit the degree of parallelism, otherwise the task may not complete.
         # FIXME: This number should be set according to the resources available.
         torch.set_num_threads(4)
         model = sentence_transformers.SentenceTransformer(model_name)  # type: ignore reportUnknownVariableType
-        SD["model"] = model  # type: ignore reportOptionalSubscript
+        sd_["model"] = model  # type: ignore reportOptionalSubscript
     else:
-        model = SD["model"]  # type: ignore reportOptionalSubscript
+        model = sd_["model"]  # type: ignore reportOptionalSubscript
 
     # Sentences are encoded by calling model.encode()
     emb = model.encode(content, normalize_embeddings=True)  # type: ignore reportUnknownVariableType


greenplumpython/experimental/embedding.py
```diff
@@ -0,0 +1,174 @@
...
+import greenplumpython as gp
+
+_vector_type = gp.type_("vector", modifier=384)
+
+
+@gp.create_function
+def _generate_embedding(content: str, model_name: str) -> _vector_type:
+    import sys
+
+    SD = globals().get("SD", sys.modules["plpy"]._SD)
```
ruxuez, line 14: `sys.modules["plpy"]._SD` this could occurs error because of dill

`ExternalRoutineException: AttributeError: module 'plpy' has no attribute '_SD'  (seg1 10.128.0.23:8001 pid=3323603)`

greenplumpython/experimental/embedding.py
```diff
@@ -0,0 +1,203 @@
...
+    This class enables users to search unstructured data based on semantic similarity and to leverage the power of
+    the vector index scan.
+    """
+
+    def __init__(self, dataframe: gp.DataFrame) -> None:
+        self._dataframe = dataframe
+
+    def create_index(self, column: str, model: str) -> gp.DataFrame:
+        """
+        Generate embeddings and create index for a column of unstructured data.
```
ruxuez, line 42:
```suggestion
        Create index for a column of unstructured data based on embeddings generated by selected model.
```

greenplumpython/experimental/embedding.py
```diff
@@ -0,0 +1,204 @@
...
+from uuid import uuid4
+
+import greenplumpython as gp
+from greenplumpython.func import FunctionExpr
+from greenplumpython.row import Row
+
+
+@gp.create_function
+def _generate_embedding(content: str, model_name: str) -> gp.type_("vector", modifier=384):  # type: ignore reportUnknownParameterType
+    import sys
```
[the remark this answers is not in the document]
ruxuez: Yes, we should definitely limit the degree of parallelism, otherwise the task won't end when the number of data increases. The number will depends on number of cores and memory, if we want a constant number  4 seems fine to me.

greenplumpython/experimental/embedding.py
```diff
@@ -0,0 +1,204 @@
...
+            END;
+            $$;
+            """,
+            has_results=False,
+        )
+        return self._dataframe
+
+    def search(self, column: str, query: Any, top_k: int) -> gp.DataFrame:
+        """
+        Searche unstructured data based on semantic similarity on embeddings.
```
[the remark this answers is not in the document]
ruxuez: Let me push it

greenplumpython/dataframe.py
```diff
@@ -1183,4 +1207,9 @@ def from_columns(cls, columns: Dict[str, Iterable[Any]], db: Database) -> "DataF
         columns_string = ",".join(
             [f'unnest({_serialize_to_expr(list(v), db=db)}) AS "{k}"' for k, v in columns.items()]
         )
-        return DataFrame(f"SELECT {columns_string}", db=db)
+        return cls(f"SELECT {columns_string}", db=db)
+
+    # Add interface here for language servers.
+    def embedding(self) -> "Embedding":
+        """Allow user to process vector operators thanks to :class:`~experimental.embedding.Embedding`."""
```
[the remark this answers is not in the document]
ruxuez: Should we remind users that they should do `import greenplumpython.experimental.embedding` otherwise they will get a `NotImplementedError` from this function when they try to do `t.embedding().create_index(column, model)`

doc/source/emb_example.rst
```diff
@@ -0,0 +1,7 @@
+.. _embedding-example:
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Generating, Indexing and Searching Embeddings
```
ruxuez, line 5: We can shorten this caption if we want it to be more concise when referenced.

greenplumpython/experimental/embedding.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+@gp.create_function
+def _generate_embedding(content: str, model_name: str) -> gp.type_("vector"):  # type: ignore reportUnknownParameterType
+    import sys
+
+    import sentence_transformers  # type: ignore reportMissingImports
+
+    _sd = globals().get("SD")
+    if globals().get("SD") is None:
```
yihong0618, line 17: if _sd is not None?

yihong0618 (commented):
LGTM 
one question, do we have test for the embeddings?

yihong0618 (approved):

ruxuez (approved):

PR #208, opened 2023-07-28
Make db a parameter of serialize()

xuebinsu:
Previously, `db` is an attribute of class `Expr` and its value
cannot be decided when init a new object. As a result, when the
`Expr` is serialized, we need to explicitly bind it to a db to
create functions or types. However, in some cases we forgot to
do that, which lead to errros.

To avoid forgotting this, this patch moves db to a parameter of
`serialize()` and pass db to it each time we call it.

This patch fixes the following bugs in `serialize()`:

1. Database is not found when trying to create a function or data
   type.
2. Star notation `*` is not serialized correctly in column fileds
   when expanding a composite type result.
3. `count(*)` returns wrong result when dataframe contains `None`.

### DIFF
--- a/tests/test_func.py
+++ b/tests/test_func.py
@@ -829,30 +829,7 @@ def test_func_in_binary_expr(db: gp.Database):
         assert row["val"] == (1 + 2) + (1 + 1)
 
 
-def test_func_after_where(db: gp.Database):
-    # fmt: off
-    rows = [(i, i,) for i in range(0, 10)]
-    # fmt: on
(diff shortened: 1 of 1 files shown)


ruxuez (commented):
LGTM. Perhaps we should add an extra test cases to cover `operators`, select conditions in brackets and with `where()` method to add a bit more complexity to the test.

yihong0618 (approved):

ruxuez (approved):

PR #209, opened 2023-07-28
Add drop when save a dataframe in database

ruxuez:


greenplumpython/dataframe.py
```diff
@@ -899,6 +900,7 @@ def save_as(
             schema: schema of the table for avoiding name conflicts.
             distribution_type: type of distribution by.
             distribution_key: distribution key.
+            drop_if_exists: bool to indicate if drop table if exists.
```
higuoxing, line 903:
```suggestion
            overwrite: boolean value to indicate whether to overwrite the table with the same name if exist.
```

What do you think of using `overwrite` as the flag name?

higuoxing (approved):

yihong0618 (approved):

PR #213, opened 2023-09-01
Set client encoding to UTF-8 for non-ASCII

xuebinsu:
When client encoding is not specified,  psycopg2 will use `latin1`.
This causes failure for non-ASCII characters to be decoded in
UTF-8 when being serialized.

This patch fixes the issue by setting the client encoding to `utf-8`


beeender (approved):

beeender (approved):
lint fails

yihong0618 (approved):
LGTM

PR #214, opened 2023-09-01
Replace gp_segment_id with UDF

xuebinsu:
Since commit 6467882a, access to gp_segment_id is explicitly disabled. (Before the PR they can be accessed just "by accident") This is because they are too limited and only valid for "regular tables".

This patch replaces gp_segment_id with UDF gp_execution_segment(). The latter can be applied on not only regular tables, but also queries. Therefore, it is more general than gp_segment_id

### DIFF
--- a/doc/source/notebooks/abalone.ipynb
+++ b/doc/source/notebooks/abalone.ipynb
@@ -30,7 +30,7 @@
   },
   {
    "cell_type": "code",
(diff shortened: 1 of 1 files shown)


ruxuez (approved):
LGTM

yihong0618 (approved):

PR #216, opened 2023-09-13
Fix missing dependency on segments

xuebinsu:
Previsously, to record dependency, we used `UPDATE` in SQL to update pg_depend. However, this only took effect on coordinator but not on segments. As a result, the embedding index cannot be drop on segments even with `DROP CASCADE`.

This patch fixes this issue by calling the C function `recordDependencyOn()` with ctypes on all segements including coordinator. As a result, `DROP CASCADE` will also clean the embedding index on segments.

### DIFF
--- a/greenplumpython/experimental/embedding.py
+++ b/greenplumpython/experimental/embedding.py
@@ -151,20 +151,19 @@ class Embedding:
                 FROM embedding_info
                 WHERE oid = '{self._dataframe._qualified_table_name}'::regclass::oid;
 
-                PERFORM 
-                    gp_execution_segment(), 
+                PERFORM
                     {_record_dependency._qualified_name_str}(
                         '{self._dataframe._qualified_table_name}'::regclass::oid,
                         '{embedding_df._qualified_table_name}'::regclass::oid
-                    )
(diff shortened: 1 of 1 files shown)


yihong0618:
can the code run well in pg? IMO, for pg there seems no segments

greenplumpython/experimental/embedding.py
```diff
@@ -120,28 +151,23 @@ def create_index(self, column: str, model: str) -> gp.DataFrame:
...
+                        '{embedding_df._qualified_table_name}'::regclass::oid
+                    );
+                IF version() LIKE '%Greenplum%' THEN
+                    PERFORM
+                        {_record_dependency._qualified_name_str}(
+                            '{self._dataframe._qualified_table_name}'::regclass::oid,
+                            '{embedding_df._qualified_table_name}'::regclass::oid
+                        )
+                    FROM gp_dist_random('gp_id');
+                END IF;
```
yihong0618, line 166: learned this thanks

yihong0618 (approved):
LGTM

ruxuez (approved):

PR #217, opened 2023-09-18
Support bulk data loading from client

xuebinsu:
Previously, to upload data from client to the database, the only way is to call `DataFrame.from_rows()` or `DataFrame.from_columns()`. In this way, data will be serialized as part of the query. When data is large, the query will become very long and can take large amount of memory.

This patch fixes the issue by uploading the data files to the database with `COPY FROM STDIN`. In this way, the query can be kept short. After the data files are copied, a UDF is called to parse the files into rows. This makes it much easier to support various data formats, inluding the unseen ones.

### DIFF
--- a/greenplumpython/db.py
+++ b/greenplumpython/db.py
@@ -51,6 +51,7 @@ class Database:
         ]  # To tell which variant of PostgreSQL is
 
     def _is_variant(self, full_name: str) -> bool:
+        assert len(full_name) > 4, "Name of the variant is expected to contain > 4 characters."
         return full_name.capitalize() in self._version
 
     def _execute(


greenplumpython/experimental/file.py
```diff
@@ -0,0 +1,79 @@
...
+@classmethod
+def _from_files(_, files: list[str], parser: NormalFunction, db: gp.Database) -> gp.DataFrame:
+    tmp_archive_name = f"tar_{uuid.uuid4().hex}"
+    tmp_archive_base = pathlib.Path("/") / "tmp" / tmp_archive_name
+    tmp_archive_base.mkdir()
+    tmp_archive_path = tmp_archive_base / f"{tmp_archive_name}.tar.gz"
+    with tarfile.open(tmp_archive_path, "w:gz") as tmp_archive:
+        for file_path in files:
+            tmp_archive.add(pathlib.Path(file_path))
+    with psycopg2.connect(db._conn.dsn, options="-c gp_session_role=utility") as util_conn:
```
beeender, line 50: The `GreenplumPython` needs to support pg as well. `gp_session_role` won't be needed for pg. I am not sure if an error will be reported if pg doesn't recognize the option,

greenplumpython/experimental/file.py
```diff
@@ -0,0 +1,79 @@
...
+
+import greenplumpython as gp
+from greenplumpython.func import NormalFunction
+
+_CHUNK_SIZE = 256 * 1024 * 1024  # Must be much < 1 GB
+
+
+@gp.create_function
+def _dump_file_chunk(tmp_archive_name: str, chunk_base64: str) -> int:
+    tmp_archive_base = pathlib.Path("/") / "tmp" / tmp_archive_name
```
beeender, line 19: Fixing the directory to `/tmp` may not be ideal. We probably need to consider give user options to choose.

`XDG_CACHE_HOME` may be preferred as the default. See https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

But this is out of the scope of this PR. If needed, it can be done in a new PR.

greenplumpython/experimental/file.py
```diff
@@ -0,0 +1,79 @@
...
+            with open(tmp_archive_path, "rb") as tmp_archive:
+                while True:
+                    chunk = tmp_archive.read(_CHUNK_SIZE)
+                    if len(chunk) == 0:
+                        break
+                    chunk_base64 = base64.b64encode(chunk)
+                    cursor.copy_expert(
+                        f"COPY {tmp_archive_name} (text_base64) FROM STDIN",
+                        io.BytesIO(chunk_base64),
+                    )
```
beeender, line 62: @higuoxing please help to take a look at the copy part.

tests/test_file.py
```diff
@@ -0,0 +1,32 @@
...
+    def parse_csv(path: str) -> list[IntAndText]:
+        import csv
+
+        with open(path) as csv_file:
+            for row in csv.DictReader(csv_file):
+                yield row
+
+    res = list(db.create_dataframe(files=[csv_path], parser=parse_csv))
+    assert len(res) == NUM_ROWS
+    assert list(next(iter(res))) == ["i", "t"]
```
beeender, line 32: I suggest to mock `_CHUNK_SIZE ` and add a multiple chunks test.

greenplumpython/experimental/file.py
```diff
@@ -0,0 +1,79 @@
...
+    tmp_archive_base = pathlib.Path("/") / "tmp" / tmp_archive_name
+    tmp_archive_path = tmp_archive_base / f"{tmp_archive_name}.tar.gz"
+    with open(tmp_archive_path, "ab") as tmp_archive:
+        tmp_archive.write(base64.b64decode(chunk_base64))
+    return 0
+
+
+@gp.create_function
+def _extract_files(tmp_archive_name: str) -> list[str]:
+    tmp_archive_base = pathlib.Path("/") / "tmp" / tmp_archive_name
```
beeender, line 28: An alternative solution could be something like:

```
COPY base64decode(xxx) to 'cat >> /tmp/xxx.tar.gz`
```

The current one is OK.

beeender: Uh, OK. `COPY` has a `NEWLINE` option, but it needs one of `CR`/`LF`/`CRLF`. 
This may not work well without further hacks.

yihong0618:
server side data, may be not in client side

yihong0618 (approved):
LGTM since I had looked into this code yesterday

tests/test_file.py
```diff
@@ -0,0 +1,43 @@
...
+
+    df_from_gp = pd.DataFrame.from_records(
+        iter(db.create_dataframe(files=[csv_path], parser=parse_csv))
+    )
+    assert df.equals(df_from_gp)
+
+
+def test_csv_multi_chunks(db: gp.Database):
+    # Set the chunk size to be 3 bytes (< size of int in C)
+    # so that data is guaranteed to be splitted into multiple chunks.
```
beeender, line 38: Does below work?

```
with patch("greenplumpython.experimental.file._CHUNK_SIZE", 3):
    test_csv_single_chunk(db)
```

beeender (approved):

PR #218, opened 2023-09-19
Support installing Python packages from client

xuebinsu:
On servers that have no Internet access, it is hard to install the required Python packages for a UDF.

To mitigate this issue, this patch enables installing packages from the client side. This works as follows:

1. Download the required packages to a directory on client with `pip download`
2. Archive and upload the downloaded packages to server
3. Install the uploaded packages with `pip install --find-links`.

Please note that this only install the packages on a single server host. To make the packages available to other hosts, an NFS on the server side might be helpful.

This patch also fixes a bug to enable UDF arguments to use SQL keywords as the names by double-quoting them.

### DIFF
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -66,9 +66,7 @@ def pip_show(pkg_name: str) -> str:
 
 @gp.create_function
 def sys_path() -> list[str]:
-    yield sys.flags
-    for p in sys.path:
-        yield p
+    return sys.path
 
 
 def test_intall_pacakges(db: gp.Database):


beeender (approved):
with some minor comments

greenplumpython/experimental/file.py
```diff
@@ -69,11 +71,75 @@ def _from_files(_, files: list[str], parser: NormalFunction, db: gp.Database) ->
...
+        "pip",
+        "install",
+        "--no-index",
+        "--requirement",
+        "/dev/stdin",
+        "--find-links",
+        cache_dir,
+    ]
+    try:
+        output = sp.check_output(cmd, text=True, stderr=sp.STDOUT, input=requirements)
```
yihong0618, line 109: here, can we use something as import instead of subprocess?
```
def install(package):
    if hasattr(pip, 'main'):
        pip.main(['install', package])
    else:
        pip._internal.main(['install', package])
```

greenplumpython/experimental/file.py
```diff
@@ -69,11 +71,75 @@ def _from_files(_, files: list[str], parser: NormalFunction, db: gp.Database) ->
...
+    cache_dir.mkdir(parents=True)
+    cmd = [
+        sys.executable,
+        "-m",
+        "pip",
+        "download",
+        "--requirement",
+        "/dev/stdin",
+        "--dest",
+        cache_dir,
```
yihong0618, line 129: ditto

ruxuez (approved):
Well done! LGTM!

PR #220, opened 2023-10-10
Migrate to monorepo for CI

xuebinsu:


liuxueyang:
This PR should be merged after the corresponding PR in gp-extensions-ci is merged.

higuoxing:
> This PR should be merged after the corresponding PR in gp-extensions-ci is merged.

+1.

higuoxing (approved):

concourse/test.sh
```diff
@@ -0,0 +1,39 @@
...
+}
+
+function _main() {
+    install_plpython3
+
+    source "$CI_REPO_DIR/common/entry_common.sh"
+    sudo passwd --delete gpadmin  # for `su gpadmin` in start_gpdb
+    echo 'alias su="sudo su"' >>"$HOME/.bashrc"
+    source "$HOME/.bashrc"  # for `alias su`
+    start_gpdb
```
beeender, line 22: why not putting `start_gpdb` to `pre_test.sh` like others? https://github.com/pivotal/advanced-password-check/blob/gpdb/concourse/pre_test.sh
- `pre_test.sh` runs with root permission, no need to hack the `su`/`sudo` anymore
- it will add `source` to `.bashrc`, so everything will be set up when gpadmin login.
- Ideally, things in `pre_test.sh` do setup things, and runs only once. So `test.sh` can be run again when hijack to the concourse.

concourse/test.sh
```diff
@@ -0,0 +1,55 @@
...
+        popd
+    fi
+}
+
+start_gpdb_as_gpadmin() {
+    if ! gpstate; then
+        source "$CI_REPO_DIR/common/entry_common.sh"
+        sudo passwd --delete gpadmin  # for `su gpadmin` in start_gpdb
+        if [[ "$(source /etc/os-release && echo "$ID")" == "ubuntu" ]]; then
+            su () {
```
liuxueyang, line 20: I don't like this kind of hack. 🤣

concourse/test.sh
```diff
@@ -0,0 +1,55 @@
...
+                sudo "su" "$@"
+            }
+        fi
+        set +o nounset
+        start_gpdb
+        source "$HOME/.bashrc"  # for gpdemo-env.sh
```
liuxueyang, line 26: I guess ~/.bashrc is sourced automatically before the `test.sh` is run?

liuxueyang (approved):
There are some hacks in the script which are not necessary in my opinion. But it is OK it works well. 😆

concourse/test.sh
beeender: Any special reason why not just follow the convention to have a `pre_test.sh`to start gpdb there?

beeender: OK. saw https://github.com/pivotal/gp-extensions-ci/pull/31#issuecomment-1780352588

beeender (approved):
> There are some hacks in the script which are not necessary in my opinion. But it is OK if it works well. 😆

Agreed.

PR #221, opened 2023-10-19
Fix error when `/tmp` is a symlink

xuebinsu:
On macOS, `/tmp` is a symlink to `/private/tmp`. When creating the tarball, the symlink will be resolved, making the file paths on server are different from those on client. This results in a `FileNotFoundError`.

This patch fixes the issue by resolving the local path to the real one if it is a symlink.

### DIFF
--- a/greenplumpython/experimental/file.py
+++ b/greenplumpython/experimental/file.py
@@ -118,7 +118,7 @@ def _from_files(_, files: list[str], parser: NormalFunction, db: gp.Database) ->
             lambda: parser(_extract_files(server_tmp_dir, tmp_dir_handle, "files")),
             expand=len(result_members) == 0,
         )
-        # _remove_tmp_dir(util_conn, db, tmp_dir_handle)  # Cannot remove now since the returning DataFrame depends on it.
+        # _remove_tmp_dir(util_conn, db, tmp_dir_handle)  # Cannot remove now since the returned DataFrame depends on it.
         return df
 
 


higuoxing (approved):

greenplumpython/experimental/file.py
```diff
@@ -121,7 +121,7 @@ def _install_on_server(pkg_dir: str, requirements: str) -> str:
 def _install_packages(db: gp.Database, requirements: str):
     tmp_archive_name = f"tar_{uuid.uuid4().hex}"
     # FIXME: Windows client is not supported yet.
-    local_dir = pathlib.Path("/") / "tmp" / tmp_archive_name / "pip"
+    local_dir = (pathlib.Path("/") / "tmp").resolve() / tmp_archive_name / "pip"
```
beeender, line 124: Instead of hard code `/tmp` path, libs like `tempfile` is preferred to be used. https://docs.python.org/3/library/tempfile.html

`/tmp` is a convention, and users can actually set the temp directory to something else.

yihong0618 (approved):

greenplumpython/experimental/file.py
beeender: > If, as you said, users can actually set the temp directory to something else, then /tmp is not a hard-coded path, but can be set to point to anywhere by user.

Yes, see https://www.baeldung.com/linux/change-tmp-directory-path

> Based on my testing , tempfile also creates files and directories under /tmp, the same as the previous scheme.

Not quite sure about the question.

> tempfile will delete the files upon exiting the context. Would this make debugging harder?

There is argument for this 
```
class tempfile.TemporaryDirectory(suffix=None, prefix=None, dir=None, ignore_cleanup_errors=False, *, delete=True)
The delete parameter can be used to disable cleanup of the directory tree upon exiting the context. While it may seem unusual for a context manager to disable the action taken when exiting the context, it can be useful during debugging or when you need your cleanup behavior to be conditional based on other logic.


[¶](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory)
```

beeender: > To confirm, are you OK with using directory under /tmp for storing the files here?

I prefer to have a subdir to store *temporary* files, for example, store tmp files in `/tmp/sOmEUuid/` actually, won't this be the same with previouse code? previously, `tmp_archive_name` is a uuid and it is the subdir name in `/tmp`. `tempfile.TemporaryDirectory()` does the same thing.

> Why tempfile is preferred over the previous scheme?

- it handles all platforms temp directory
- it handles customized `TMPDIR`

beeender:
Is this ready for the next round of review? I saw CI failed.

beeender (approved):

PR #224, opened 2023-11-02
Support more index types besides ivfflat

xuebinsu:
Previously, we only support indexing embeddings using the `ivfflat`
access method in [pgvector](https://github.com/pgvector/pgvector).

Recently, a new access method `hnsw` has been added to pgvector.
`hnsw` is believed to be more performant and accurate than `ivfflat`.
To allow for more flexibility, we add a new parameter `method` to
allow user to choose which access method to use when creating index.

Also, a new parameter `embedding_dimension` is added to support more
models, since dimension is required for pgvector to create index.

A new test case for embeddings is added in `tests/test_embedding.py`.

To support `set allow_system_table_mods = on;`, Postgres is upgraded
from 12 to 13 on CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,11 +16,11 @@ jobs:
       fail-fast: true
       matrix:
         python-version: ["3.9", "3.11"]
-        server: ["postgres12-python39", "postgres12-python311"]
+        server: ["postgres13-python39", "postgres13-python311"]
         include:
-          - server: "postgres12-python39"
+          - server: "postgres13-python39"
             server-python-version: "3.9"
-          - server: "postgres12-python311"
+          - server: "postgres13-python311"
             server-python-version: "3.11"
   
     steps:
--- a/server/postgres13-python311.Dockerfile
+++ b/server/postgres13-python311.Dockerfile
@@ -1,4 +1,4 @@
-FROM postgres:12-bookworm
+FROM postgres:13-bookworm
 
 COPY build.sh initdb.sh requirements.sh /tmp/
 RUN bash /tmp/build.sh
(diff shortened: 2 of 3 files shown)


beeender (commented):
Can we have some tests for embedding? Is there any blockers for testing it on CI?

greenplumpython/experimental/embedding.py
```diff
@@ -105,17 +113,17 @@ def create_index(self, column: str, model_name: str) -> gp.DataFrame:
...
-            )
+        if embedding_dimension is None:
+            try:
+                import sentence_transformers  # type: ignore reportMissingImports
+
+                model = sentence_transformers.SentenceTransformer(model_name)  # type: ignore reportUnknownVariableType
+                embedding_dimension: int = model[1].word_embedding_dimension  # From models.Pooling
+            except:
+                raise NotImplementedError(
+                    "Model '{model_name}' doesn't provide embedding dimension information"
```
yihong0618, line 125: maybe import error here

greenplumpython/experimental/embedding.py
```diff
@@ -136,8 +144,10 @@ def create_index(self, column: str, model_name: str) -> gp.DataFrame:
                 distribution_type="hash",
             )
             .check_unique(self._dataframe.unique_key)
-            .create_index(columns={embedding_col_name}, method="ivfflat")
         )
+        if method is not None:
+            assert method == "hnsw" or method == "ivfflat"
```
yihong0618, line 149: since there may be more method , `assert method in ["ivfflat", "hnsw"]` may be bettor?

greenplumpython/experimental/embedding.py
yihong0618: make sense

beeender (approved):

yihong0618 (approved):
LGTM

PR #226, opened 2023-12-05
Fix missing unique key when searching embeddings

xuebinsu:
Unique key is required when searching embeddings to join the embedding table and the original data table before returning the results. Previously, search on a dataframe that created from an existing table in database failed due to lacking of unique key in the dataframe.

This patch fixes the issue by recording the unique key when `create_index()` in `pg_class` so that the info can be read when `search()`.

### DIFF
--- a/tests/test_embedding.py
+++ b/tests/test_embedding.py
@@ -52,7 +52,7 @@ def test_embedding_multi_col_unique(db: gp.Database):
         )
         .check_unique(columns={"id", "id2"})
     )
-    t.embedding().create_index(column="content", model_name="all-MiniLM-L6-v2")
+    t = t.embedding().create_index(column="content", model_name="all-MiniLM-L6-v2")
     print(
         "reloptions =",
         db._execute(


tests/test_embedding.py
```diff
@@ -19,8 +21,13 @@ def test_embedding_query_string(db: gp.Database):
         )
         .check_unique(columns={"id"})
```
beeender, line 22: 1. `check_unique` actually does more than *check", it creates indexes which is not obvious from the function name. Shall we consider to rename the function?
2. Need tests multi columns for `check_unique()` and `search()`. Another PR will be fine since it is not relevant to this one.

beeender:
```
[5.4.1. Check Constraints](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS)
[5.4.2. Not-Null Constraints](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-NOT-NULL)
[5.4.3. Unique Constraints](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS)

5.4.1. Check Constraints

A check constraint is the most generic constraint type. It allows you to specify that the value in a certain column must satisfy a Boolean (truth-value) expression. For instance, to require positive product prices, you could use:

CREATE TABLE products (
    product_no integer,
    name text,
    price numeric CHECK (price > 0)
);

```

Doesn't this mean `CHECK` is one kind of constrains, and `UNIQUE` is another kind of constrain?

Is `check` in `check_unique` a verb? Or do I miss anything?

beeender (approved):

tests/test_embedding.py
```diff
@@ -1,15 +1,24 @@
...
 import greenplumpython as gp
+import greenplumpython.experimental.embedding as _
 from tests import db
 
 
+def search_embeddings(t: gp.DataFrame):
+    results = t.embedding().search(column="content", query="apple", top_k=1)
+    assert len(list(results)) == 1
+    for row in results:
+        assert row["content"] == "I like eating apples."
```
yihong0618, line 12: since only one item here no need to for loop?

tests/test_embedding.py
```diff
@@ -19,8 +28,35 @@ def test_embedding_query_string(db: gp.Database):
...
+        .check_unique(columns={"id", "id2"})
+    )
+    t.embedding().create_index(column="content", model_name="all-MiniLM-L6-v2")
+    print(
+        "reloptions =",
+        db._execute(
+            f"SELECT reloptions FROM pg_class WHERE oid = '{t._qualified_table_name}'::regclass"
+        ),
+    )
+    search_embeddings(t)
```
yihong0618, line 62: is this test always pass? seems we have no assert here

yihong0618 (approved):
LGTM

PR #231, opened 2024-01-08
Support Existing Table Column Information

ruxuez:
So far, we don't have any information about the DataFrame's 
columns and their types. This patch supports retrieving column 
information of an existing table in the database. 
Moreover, adds a table check when creating a DataFrame from table, 
so that the user can catch the name error early.

### DIFF
--- a/greenplumpython/dataframe.py
+++ b/greenplumpython/dataframe.py
@@ -973,21 +973,6 @@ class DataFrame:
             if drop_if_exists
             else ""
         )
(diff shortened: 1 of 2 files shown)


ruxuez:
> How about exposing this functionality as `DataFrame.describe()`, so that this can be done lazily, i.e. only when user needs the info?

Do you mean for now only do it lazily?

yihong0618 (approved):

